PR #10, opened 2023-07-05
Add Github CI workflow for every push commit

myrrc:


Makefile
```diff
@@ -102,22 +102,42 @@ prepare-version: version.txt
...
-.PHONY: build-deb-package
-build-deb-package: prepare-changelog
-	@echo 'Building debian package'
+lint: black isort #pylint mypy bandit
 
-	rm -f pyproject.toml  # for forcing usage of setup.py
-	cd debian && \
-		debuild --check-dirname-level 0 --preserve-env --no-lintian --no-tgz-check -uc -us
+black: install-deps
+	black --check src/ tests/
```
dstaroff, line 112: Let's pass `-S/--skip-string-normalization` option here to skip string normalization

Makefile
```diff
@@ -102,22 +102,42 @@ prepare-version: version.txt
...
+isort: install-deps
+	isort src/ tests/
```
dstaroff, line 114: And let's add explicit `--project chtools` option too

pyproject.toml
```diff
@@ -1,55 +1,51 @@
 [build-system]
-requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
+requires = ["flit_core >=3.2,<4"]
+build-backend = "flit_core.buildapi"
 
-[tool.poetry]
-name = "ch-tools"
+[project]
+name = "chtools"
```
dstaroff, line 6: Is there a reason to eliminate the hyphen in the package name?

dstaroff (changes requested):

pyproject.toml
myrrc: Yes, flit doesn't accept project name otherwise. I guess we could change src/chtools to src/ch-tools as an option

myrrc: The reason is that flit's package name must be equal to directory name in src/

dstaroff (approved):

PR #12, opened 2023-07-06
fail critically after 2 days of missing backups

fdondi-doublecloud:
https://st.yandex-team.ru/ORION-3142

Monitoring currently does not fail if backups are not created at all, and thus do not count as "failed". 
Fail if no backups are present for more than 2 days.


myrrc (approved):
LGTM

PR #20, opened 2023-07-13
Build deb-package workflow

dstaroff:


Makefile
```diff
@@ -4,11 +4,47 @@ PREFIX=/opt/yandex/mdb-ch-tools
...
-install: install-symlinks install-bash-completions install-logrotate ;
+install: install-python-package install-symlinks install-bash-completions configure-logs ;
 
 .PHONY: uninstall
 uninstall: uninstall-python-package uninstall-symlinks uninstall-bash-completions uninstall-logrotate ;
 
+.PHONY: install-python-package
+install-python-package: build-python-package
+	@echo 'Installing mdb-ch-tools'
+
```
myrrc, line 15: Why can't you rely on `flit install` here?

.github/workflows/build-deb-package.yml
```diff
@@ -0,0 +1,69 @@
...
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  build-deb-package:
+    strategy:
+      matrix:
+        target:
+          - { python: "3.6.15", ubuntu: 20.04 }
+          - { python: "3.10.12", ubuntu: 22.04 }
```
myrrc, line 16: Why do we need a separate pipeline for that? We can reuse existing workflow and build a .deb package every commit like we do with .whl files

Makefile
```diff
@@ -78,29 +114,43 @@ prepare-version: version.txt
...
 	@echo "--------------------------------------------------------------------------------"
 	@echo ""
 	@echo "Debian package build targets:"
 	@echo "  install                    Install 'mdb-ch-tools' debian package"
 	@echo "  uninstall                  Uninstall 'mdb-ch-tools' debian package"
 	@echo ""
+	@echo "  install-python-package     Install 'ch-tools' python package"
 	@echo "  uninstall-python-package   Uninstall 'ch-tools' python package"
 	@echo "  install-symlinks           Install symlinks to /usr/bin/"
 	@echo "  uninstall-symlinks         Uninstall symlinks from /usr/bin/"
```
myrrc, line 150: Btw `flit install` also handles symlinking to executables, so you wouldn't need a separate step for that

myrrc (changes requested):

pyproject.toml
```diff
@@ -4,9 +4,8 @@ build-backend = "flit_core.buildapi"
 
 [project]
 name = "chtools"
```
Alex-Burmak, line 6: it doesn't relate to this PR but I propose to rename it to `ch-tools`.

It should be possible to resolve issue with import name with the following directive:
```
[tool.flit.module]
name = "chtools"
```
https://flit.pypa.io/en/latest/pyproject_toml.html#module-section

myrrc: I'll update this once all pipelines are working.

.github/workflows/workflow.yml
```diff
@@ -104,3 +114,18 @@ jobs:
       if: always()
       with:
         report_paths: 'tests/reports/*.xml'
+
+    - name: build debian package
+      # language=sh
+      run: |
+        sudo apt-get update -q -y
+        sudo apt-get install -q -y git python3 python3-venv python3-pip debhelper devscripts
+        mv dist/*.tar.gz chtools.tar.gz
```
myrrc, line 123: python and pip are already present, see `setup-python` version above

.github/workflows/workflow.yml
```diff
@@ -104,3 +114,18 @@ jobs:
...
+        make clean build-deb-package
+
+    - name: upload deb package artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: mdb-ch-tools_py${{ matrix.target.python }}.deb
+        path: out/mdb-ch-tools*.deb
```
myrrc, line 130: Let's also remove mdb here

myrrc (changes requested):

.github/workflows/workflow.yml
dstaroff: When I tested this workflow with act, python3* packages weren't present. Maybe the reason is in medium container image instead of a full one, but I suggest being sure on these packages

.github/workflows/build-deb-package.yml
dstaroff: Fixed

Makefile
dstaroff: These executables are installed on host's `/usr/bin`, not in `$DESTDIR/usr/bin`

Makefile
dstaroff: Now we use artifact from built with flit

.github/workflows/build-deb-package.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: upload amd64 artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: mdb-ch-tools_py${{ matrix.target.python }}_amd64.deb
+          path: ../mdb-ch-tools*.deb
+          if-no-files-found: error
+
+      - name: upload arm64 artifacts
+        uses: actions/upload-artifact@v3
```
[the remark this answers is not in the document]
dstaroff: I'm pretty sure that we don't need this at the moment

.github/workflows/workflow.yml
dstaroff: Fixed

.github/workflows/build-deb-package.yml
```diff
@@ -0,0 +1,69 @@
...
+          run: |
+            sed -i -E -e "s/amd64|arm64/arm64/g" debian/control debian/rules
+            make build-deb-package
+            mv ../mdb-ch-tools*.deb /artifacts/mdb-ch-tools_py${{ matrix.target.python }}_arm64.deb
+            echo "Produced artifact: `ls /artifacts/`"
+
+      - name: upload amd64 artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: mdb-ch-tools_py${{ matrix.target.python }}_amd64.deb
```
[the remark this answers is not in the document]
dstaroff: Fixed

.github/workflows/workflow.yml
```diff
@@ -104,3 +114,29 @@ jobs:
       if: always()
       with:
         report_paths: 'tests/reports/*.xml'
+
+    - name: build debian package
+      # language=sh
```
myrrc, line 119: `if: ${{ matrix.clickhouse == 'latest' }}`

myrrc (changes requested):

myrrc (approved):

PR #32, opened 2023-08-03
Try to resolve locale before running Click

dstaroff:
On some systems, misconfigured `LC_ALL` and `LANG` env variables breaks Click with errors like:
```
Traceback (most recent call last):
  File "/usr/bin/chadmin", line 8, in <module>
    sys.exit(main())
  File "/opt/yandex/ch-tools/lib/python3.6/site-packages/ch_tools/chadmin/chadmin_cli.py", line 150, in main
    cli.main()
  File "/opt/yandex/ch-tools/lib/python3.6/site-packages/click/core.py", line 1034, in main
    _verify_python_env()
  File "/opt/yandex/ch-tools/lib/python3.6/site-packages/click/_unicodefun.py", line 100, in _verify_python_env
    raise RuntimeError("\n\n".join(extra))
RuntimeError: Click will abort further execution because Python was configured to use ASCII as encoding for the environment. Consult https://click.palletsprojects.com/unicode-support/ for mitigation steps.

This system supports the C.UTF-8 locale which is recommended. You might be able to resolve your issue by exporting the following environment variables:

    export LC_ALL=C.UTF-8
    export LANG=C.UTF-8

Click discovered that you exported a UTF-8 locale but the locale system could not pick up from it because it does not exist. The exported locale is 'en_US.UTF-8' but it is not supported.
```
This PR is about resolving locale before running Click


Makefile
```diff
@@ -157,8 +157,8 @@ mypy:
 
 .PHONY: format
 format:
-	isort .
-	black .
+	isort src tests
+	black src tests
```
myrrc, line 161: Is this change important? Are there any files that we don't want to format?

myrrc (commented):
LGTM

Makefile
dstaroff: venv

aalexfvk (approved):

PR #33, opened 2023-08-09
Use Poetry as a dependency tool

aalexfvk:
1. Use Poetry as a single tool for environment, dependency managment, building packages and publishing to PyPI.
2. Use flat source code layout instead of src-layout (like in ch-backup)
3. Related refactoring


aalexfvk:
https://github.com/yandex/ch-tools/actions/runs/5810475057
Result of running this workflow for version tag pushed

.github/workflows/workflow.yml
```diff
@@ -43,129 +89,85 @@ jobs:
...
-      if: ${{ matrix.clickhouse == 'latest' && matrix.target.ubuntu == 'latest' && startsWith(github.ref, 'refs/tags/') }}
-      run: flit publish --no-use-vcs
+          ${{ env.BUILD_PYTHON_OUTPUT_DIR }}/*.whl
+          ${{ env.BUILD_PYTHON_OUTPUT_DIR }}/*.tar.gz
+          ${{ env.BUILD_DEB_OUTPUT_DIR }}/ch-tools*.deb
+
+    - name: publish to PYPI
+      if: ${{ matrix.target.ubuntu == 'latest' && startsWith(github.ref, 'refs/tags/') }}
+      continue-on-error: true
+      run: make publish
```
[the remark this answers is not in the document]
Alex-Burmak: Let's try `clickhouse-tools`.

README.md
```diff
@@ -15,33 +15,36 @@
...
+make unit-tests PYTEST_ARGS="-k test_name"
 
 # integration tests (rebuild docker images using a .whl file)
-cd tests; behave
+make integration-tests
 
 # integration tests (do not rebuild docker images)
 # useful when you didn't change source code
-cd tests; behave -D skip_setup
+make integration-tests BEHAVE_ARGS="-D skip_setup"
```
Alex-Burmak, line 41: It would be good to unify naming of targets in `ch-tools` and `ch-backup`.

In ch-backup targets are named `test-integration` and `test-unit`.
https://github.com/yandex/ch-backup/blob/8737ac84fe13f82ddb7be37e97b5b9a89fc8e18d/Makefile#L58

I have no preference on naming schema. We can rename targets either here or in ch-backup.

Also it makes sense to unify semantic of targets. In ch-backup the target `test-integration` implies `"-D skip_setup"`. And there is a logic in makefile that avoid docker image rebuild if there are no source code changes.

pyproject.toml
```diff
@@ -1,41 +1,44 @@
...
-build-backend = "flit_core.buildapi"
+requires = ["poetry-core>=1.0.0"]
+build-backend = "poetry.core.masonry.api"
 
-[project]
+[tool.poetry]
 name = "ch-tools"
-license = { file = "LICENSE" }
-dynamic = ["version", "description"]
+version = "2.529.131159744"
```
Alex-Burmak, line 7: Shouldn't it be set dynamically using value from `version.txt`?

And target versioning schema, I suppose, should be based on git tags. Probably with the help of https://github.com/pypa/setuptools_scm. Though this is out of scope of the given PR.

pyproject.toml
```diff
@@ -67,68 +70,65 @@ classifiers = [
...
+mypy = { version = "^1.4.1", python = ">=3.10" }
+pyhamcrest = "*"
+pylint = { version = "^2.17.5", python = ">=3.10" }
+pytest = "*"
+types-pyOpenSSL = "*"
+types-python-dateutil = "*"
+types-pyyaml = "*"
+types-requests = "*"
+types-setuptools = "*"
+types-tabulate = "*"
```
Alex-Burmak, line 118: `types-*` packages don't make much sense without `mypy`. So it's worth to add `python = ">=3.10"` for `types-*` packages too. This should allow to use up-to-date versions of `pyyaml`, `boto` and `click`. At least in ch-backup it helps - https://github.com/yandex/ch-backup/pull/26/commits/8a02cad113160ddfb2e4f4f451f44e3c87071121#diff-2b4945591edfeaa4cf4d3f155e66d4b43d1bda7a55d881d5cf3107f1b05abbbc / https://github.com/yandex/ch-backup/pull/26/

Makefile
```diff
@@ -1,198 +1,336 @@
...
+		echo "Installing poetry $(POETRY_VERSION)..."
+		curl -sSL https://install.python-poetry.org | POETRY_HOME=$(POETRY_HOME) $(PYTHON) - --version $(POETRY_VERSION)
+
+		# Fix cannot "import name 'appengine' from 'urllib3.contrib'..." error 
+		# while 'poetry publish' for version poetry 1.1.15
+		# https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html#importerror-cannot-import-name-gaecontrib-from-requests-toolbelt-compat
+		$(POETRY_HOME)/venv/bin/python -m pip install urllib3==1.26.15
+	else
+		echo "Found installed poetry $$($(POETRY) --version)"				
+	fi
```
Alex-Burmak, line 79: Wouldn't it be easier to install poetry using `pip` and `venv`?

https://gist.githubusercontent.com/Alex-Burmak/ca4f748a9673748707205e4fd62752f1/raw/f6c05e6ffaab96d190b4fc5742abaad9a7b7ab93/venv.txt

It can be either the same venv as for other dependencies or dedicated venv for poetry.

Makefile
```diff
@@ -1,198 +1,336 @@
...
+install-deps: $(INSTALL_DEPS_STAMP) ;
+
+
+# Update dependencies in poetry.lock to their latest versions according to your pyproject.toml
+.PHONY: update-deps
+update-deps:
+	$(POETRY) update
+
+
+$(INSTALL_DEPS_STAMP): ensure-poetry venv pyproject.toml
```
Alex-Burmak, line 63: Shouldn't `poetry.lock` be in dependency list?

.github/workflows/workflow.yml
```diff
@@ -1,19 +1,29 @@
...
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
-    - uses: actions/setup-python@v4
-      with: { python-version: "3.11", cache: pip }
-    - run: pip install .[test]
+    - uses: actions/checkout@v3                
+    - uses: ./.github/actions/setup_dependencies
+      with:
+        python-version: "3.10.12"
```
Alex-Burmak, line 25: Wouldn't it better to pin minor version rather then patch version? (e.g. use `3.10` instead of `3.10.12`)

.github/workflows/workflow.yml
```diff
@@ -1,19 +1,29 @@
...
+  pull_request:
+    branches: [main]
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
+env:
+  BUILD_DEB_OUTPUT_DIR: out
+  BUILD_PYTHON_OUTPUT_DIR: dist
+  # The latest version supporting Python 3.6
+  POETRY_VERSION: 1.1.15
```
Alex-Burmak, line 15: Don't we want to rely on poetry version defined in Makefile and keep it only in one place?

.github/workflows/workflow.yml
```diff
@@ -43,129 +89,85 @@ jobs:
         files: tests/bake.hcl
         push: true
 
+
   build:
-    needs: lint
+    needs: tests
```
Alex-Burmak, line 94: It looks a bit confusing that `build` goes after `tests`. Probably `release` would be a better job name.

.github/workflows/workflow.yml
```diff
@@ -43,129 +89,85 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
         target:
           - {python: "3.6.15", ubuntu: "20.04"}
           - {python: "3.10.12", ubuntu: "latest"}
         # a copy-paste of the above as github CI can't use env context in matrices
```
Alex-Burmak, line 101: the comment looks outdated

README.md
```diff
@@ -15,33 +15,36 @@
...
-pip install .[test]
-flit build --no-use-vcs
-flit install
+sudo make install-poetry
+# or to install in user's home directory
+make POETRY_HOME=~/opt/poetry install-poetry
+
+# For building deb packages
+sudo make prepare-build-deb
+sudo make build-deb-package
```
Alex-Burmak, line 27: I think it's better to place instructions on building deb package after ones on running tests.

And it would be great if these commands work without `sudo`.

pyproject.toml
aalexfvk: Yes it is updated dynamically [here](https://github.com/yandex/ch-tools/blob/3cc61b25704bbc71145e23410749fa61d865813b/Makefile#L279). I accedintally uploaded it after testing bulding deb packages. As for schema and other refinments, I use the current one and it can be changed in the next PRs.

pyproject.toml
aalexfvk: Agree. Did it.

Makefile
aalexfvk: Yes it is quite possible. I did it via pip in a separate venv because mixing project's and poetry's dependencies in one environment is a bad idea.

Makefile
aalexfvk: Yes, it is more robust. I also did some refinements.

.github/workflows/workflow.yml
aalexfvk: Done

.github/workflows/workflow.yml
aalexfvk: Agree. I left POETRY_VERSION only in makefile.

.github/workflows/workflow.yml
aalexfvk: Done

README.md
aalexfvk: I did it. And yes, now it can work without sudo.

README.md
aalexfvk: Agree. I changed test target names like in ch-backup and added implicit `-D skip_setup` . As for other changes (manipulation of inetgration testing environments) I'd propose to to add it in a separate PR.

.github/workflows/workflow.yml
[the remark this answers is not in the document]
aalexfvk: I changed it. I had to change `name` in `pyproject.toml` and in several other places.
And for [workflow](https://github.com/yandex/ch-tools/actions/runs/5834257396) it's working https://pypi.org/project/clickhouse-tools/.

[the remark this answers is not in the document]
aalexfvk: `ch-tools` remains as part of different addresses email, repo etc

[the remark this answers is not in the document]
Alex-Burmak: Great! clickhouse-tools / ch-tools eventually get available on pypi.

And I think it's OK to not rename everything from ch-tools to clickhouse-tools. At least, do not rush with it.

Makefile
```diff
@@ -1,198 +1,360 @@
...
-	)
+		rm -f $(DESTDIR)/etc/logrotate.d/$(bin);)
+
 
 .PHONY: prepare-changelog
 prepare-changelog: prepare-version
-	@echo 'Bumping version into Debian package changelog'
-	DEBFULLNAME="Yandex LLC" DEBEMAIL="ch-tools@yandex-team.ru" dch --force-bad-version --distribution stable -v `cat version.txt` Autobuild
+	echo 'Bumping version into Debian package changelog'
+	DEBFULLNAME="Yandex LLC" DEBEMAIL="clickhouse-tools@yandex-team.ru" dch --force-bad-version --distribution stable -v $$(cat $(VERSION_FILE)) Autobuild
```
Alex-Burmak, line 271: Let's keep `ch-tools@yandex-team.ru` for now. `clickhouse-tools@yandex-team.ru` is nonexistence email.

Alex-Burmak (approved):

Makefile
aalexfvk: The result of IDE auto-replace.
Thanks for the thorough review.

PR #43, opened 2023-08-21
Change debian/control "Replaces" clause

aalexfvk:


debian/control
```diff
@@ -21,4 +21,4 @@ X-Python3-Version: >= 3.6
 Package: clickhouse-tools
 Architecture: any
 Description: A set of tools for administration and diagnostics of ClickHouse DBMS.
-Replaces: mdb-ch-tools
```
Alex-Burmak, line 24: It's better to keep`Replaces: mdb-ch-tools`.

But I don't known what exact syntax should be used.

```
Replaces: mdb-ch-tools,ch-tools
```

or

```
Replaces: mdb-ch-tools
Replaces: ch-tools
```

aalexfvk: Yes it is allowable `Replaces: package-list`

Alex-Burmak (approved):

PR #57, opened 2023-08-28
Make ch_tools/__init__.py fully generated and add missing dependencies between makefile targets

Alex-Burmak:
The number of dirty files after running `make test-integration` is reduced to just `pyproject.toml`.
```
$ gmake test-integration
$ git status
On branch makefile_deps
Your branch is ahead of 'origin/main' by 1 commit.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   pyproject.toml
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,47 +122,47 @@ lint: isort black codespell ruff pylint mypy
 
 
 .PHONY: isort
-isort: install-deps
+isort: generate-version install-deps
 	$(POETRY) run isort --check --diff $(SRC_DIR) $(TESTS_DIR)
 
 
 .PHONY: black
-black: install-deps
+black: generate-version install-deps
 	$(POETRY) run black --check --diff $(SRC_DIR) $(TESTS_DIR)
 
 
 .PHONY: codespell
-codespell: install-deps
+codespell: generate-version install-deps
 	$(POETRY) run codespell
 
 .PHONY: fix-codespell-errors
-fix-codespell-errors: install-deps
+fix-codespell-errors: generate-version install-deps
 	$(POETRY) run codespell -w
 
 
 .PHONY: ruff
-ruff: install-deps
+ruff: generate-version install-deps
 	$(POETRY) run ruff $(SRC_DIR) $(TESTS_DIR)
 
 
 .PHONY: pylint
-pylint: install-deps
+pylint: generate-version install-deps
 	$(POETRY) run pylint $(SRC_DIR)
 
 
 .PHONY: mypy
-mypy: install-deps
+mypy: generate-version install-deps
 	$(POETRY) run mypy --python-version=3.6 $(SRC_DIR) $(TESTS_DIR)
 
 
 .PHONY: format
-format: install-deps
+format: generate-version install-deps
 	$(POETRY) run isort $(SRC_DIR) $(TESTS_DIR)
 	$(POETRY) run black $(SRC_DIR) $(TESTS_DIR)
 
 
 .PHONY: test-unit
-test-unit: install-deps
+test-unit: generate-version install-deps
 	$(POETRY) run $(PYTHON) -m pytest $(PYTEST_ARGS) $(TESTS_DIR)/unit
 
 
@@ -201,7 +201,7 @@ install-python-package: build-python-packages
 
 
 .PHONY: build-python-packages
-build-python-packages: prepare-version clean-dist
+build-python-packages: prepare-pyproject-file clean-dist
 	echo 'Building python packages...'
 	$(POETRY) build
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -342,7 +345,6 @@ help:
 	echo "  format                     Re-format source code to conform style settings enforced by"
 	echo "                             isort and black tools."
 	echo "  prepare-changelog          Add an autobuild version entity to changelog"
-	echo "  prepare-version            Update version based on latest commit"
```
Alex-Burmak, line 345: Replaced with `generate-version` and it's internal target. It's better to not add it to the help.

Makefile
```diff
@@ -271,25 +271,27 @@ uninstall-logrotate:
...
-	# Replace version in $(SRC_DIR)/__init__.py
-	sed -i "s/__version__ = \"[0-9\.]\+\"/__version__ = \"$${VERSION}\"/g" $(SRC_DIR)/__init__.py
+.PHONY: prepare-pyproject-file
+prepare-pyproject-file: generate-version
 	# Replace version in pyproject.toml
-	$(POETRY) version $${VERSION}
-
+	VERSION=$$(cat ${VERSION_FILE}) && \
+	$(POETRY) version $${VERSION} && \
 	echo "Version: $${VERSION}"
```
Alex-Burmak, line 284: Added `&& \` to not depend on `.ONESHELL:` and improve portability.

Makefile
```diff
@@ -122,47 +122,47 @@ lint: isort black codespell ruff pylint mypy
 
 
 .PHONY: isort
-isort: install-deps
+isort: generate-version install-deps
```
aalexfvk, line 125: Why do we need a generated version for this actions (not for building) ?

Makefile
```diff
@@ -319,6 +321,7 @@ clean: clean_debuild
 	rm -rf $(BUILD_PYTHON_OUTPUT_DIR)
 	rm -rf $(VENV_DIR)
 	rm -rf $(INSTALL_DEPS_STAMP)
+	rm -rf $(VERSION_FILE) $(SRC_DIR)/__init__.py
```
aalexfvk, line 324: Why can't we keep `$(SRC_DIR)/__init__.py` ?

Makefile
Alex-Burmak: `ch_tools/__init__.py` is a part of clickhouse tools source code and some linters don't work without this file.

```
$ gmake pylint
************* Module monrun_checks_keeper.keeper_commands
ch_tools/monrun_checks_keeper/keeper_commands.py:11:0: E0401: Unable to import 'ch_tools.common.clickhouse.config' (import-error)
ch_tools/monrun_checks_keeper/keeper_commands.py:12:0: E0401: Unable to import 'ch_tools.common.result' (import-error)
************* Module monrun_checks_keeper.main
ch_tools/monrun_checks_keeper/main.py:9:0: E0401: Unable to import 'ch_tools' (import-error)
ch_tools/monrun_checks_keeper/main.py:10:0: E0401: Unable to import 'ch_tools.common.cli.context_settings' (import-error)
ch_tools/monrun_checks_keeper/main.py:11:0: E0401: Unable to import 'ch_tools.common.cli.locale_resolver' (import-error)
ch_tools/monrun_checks_keeper/main.py:12:0: E0401: Unable to import 'ch_tools.common.result' (import-error)
ch_tools/monrun_checks_keeper/main.py:13:0: E0401: Unable to import 'ch_tools.monrun_checks_keeper.keeper_commands' (import-error)
ch_tools/monrun_checks_keeper/main.py:24:0: E0401: Unable to import 'ch_tools.monrun_checks_keeper.status' (import-error)
...
$ gmake mypy
ch_tools/common/clickhouse/config/__init__.py: error: Source file found twice under different module names: "common.clickhouse.config" and "ch_tools.common.clickhouse.config"
ch_tools/common/clickhouse/config/__init__.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
ch_tools/common/clickhouse/config/__init__.py: note: Common resolutions include: a) adding `__init__.py` somewhere, b) using `--explicit-package-bases` or adjusting MYPYPATH
Found 1 error in 1 file (errors prevented further checking)
gmake: *** [Makefile:155: mypy] Error 2
```

I added `generate-version` as a dependency for all linters for consistency. Though I'm not sure what option is better - add the dependency for all linters or only for `mypy` and `pylint`.

Makefile
Alex-Burmak: In `clean` target? From my point of view, `clean` should remove all build artifacts, clean up side effects if any (e.g. destroy docker containers created in integration tests) and return working directory to the state before executing build or test targets.

If the question about removing  `$(SRC_DIR)/__init__.py` from git in overall, it also relates to clean working directory. It's better to keep manually created files and generated ones separated to avoid committing unintentional changes. Sometimes generated changed are committed to vcs, but it doesn't seem like a good solution for `__init__.py` and `pyproject.toml`. Otherwise the changes in these files should be included in every commit because every commit affects generated version value in these files.

aalexfvk: In my opinion, we should not remove `__init__.py`. It's confusing and may be unexpected when you are doing something without using makefile after such a cleaning. Moreover this file might contain some useful code in the future. 
And the version in `pyptoject.toml` remains.
I can suggest replacing versions in this files to default values `0.0.0` in `clean` target, or by pre-commit hook. For me it is more clear.

Alex-Burmak: I'm a bit confusing because I'm always thinking that clean build is import property of any software project and that running build or tests from clean working copy shouldn't produce any changes in vcs controlled files.

Though I'm agree that PR doesn't address the issue completely and we also generate changes in pyproject.toml during the build. 
Will close the PR for now as it's incomplete work right now.

PR #60, opened 2023-08-30
monrun ch-backup check: replace datetime.strptime to dateutil.parse to parse timezone with column

yahor-s:
Fix datetime parse exception on restored cluster:
`time data '2023-08-29T02:43:53.953465+03:00' does not match format '%Y-%m-%dT%H:%M:%S.%f%z'`


yahor-s:
@Alex-Burmak @MedvedewEM 
I can't assign you to reviewers, bump with comment

PR #61, opened 2023-09-01
Fix user checking

aalexfvk:
After running highstate on CH host via `pssh-sh` state running `/usr/local/yandex/ch_wait_started.py` hangs up to timeout.
Because such a command does not work
```
sudo salt-call cmd.run "timeout 5 sudo -u monitor /usr/bin/ch-monitoring ping"
[ERROR   ] Command 'timeout' failed with return code: 1
[ERROR   ] stdout: Wrong current user: root
[ERROR   ] retcode: 1
[ERROR   ] Command 'timeout' failed with return code: 1
[ERROR   ] output: Wrong current user: root
```
Or without `salt-call`
```
sudo timeout 5 sudo -u monitor /usr/bin/ch-monitoring ping
Wrong current user: root
```

The point is that `getpass.getuser()` function relies on environment variables (`LOGNAME`, `USER`, etc) but nested sudo  does not update it:
```
$ sudo -u monitor printenv USER
monitor
$ sudo sudo -u monitor printenv USER
root
$ sudo -u alexfvk sudo -u monitor printenv USER
alexfvk
```
I propose to extract name for effective uid and not to rely on environment variables.
After fix:
```
$ timeout 5 sudo /usr/bin/ch-monitoring ping
0;OK
$ timeout 5 sudo -u monitor /usr/bin/ch-monitoring ping
0;OK
$ sudo timeout 5 sudo -u monitor /usr/bin/ch-monitoring ping
0;OK
$ sudo -u alexfvk timeout 5 sudo -u monitor /usr/bin/ch-monitoring ping
0;OK
$ sudo -u alexfvk timeout 5 sudo /usr/bin/ch-monitoring ping
0;OK
$ sudo -u alexfvk timeout 5 /usr/bin/ch-monitoring ping
Wrong current user: alexfvk
$ timeout 5 /usr/bin/ch-monitoring ping
Wrong current user: alexfvk
```

### DIFF
--- a/ch_tools/monrun_checks/main.py
+++ b/ch_tools/monrun_checks/main.py
@@ -1,4 +1,3 @@
-import getpass
 import logging
 import os
 import pwd
@@ -154,9 +153,11 @@ def main():
 
 
 def check_current_user():
-    user = getpass.getuser()
+    euid = os.geteuid()
+    user = pwd.getpwuid(euid).pw_name
+
     if user != DEFAULT_USER:
-        if os.geteuid() != 0:
+        if euid != 0:
             print(f"Wrong current user: {user}", file=sys.stderr)
             sys.exit(1)
         else:


ianton-ru:
If I remember correct the reason was that when `ch-monitoring` starter from `root`, it creates some artifacts like log files with owner `root:root`. And later when it starts from user `monitor` it fails with access denied error. So `sudo /usr/bin/ch-monitoring ping` must stops with `Wrong current user: root`.

ianton-ru:
PS. May be behavior was changed since that time and now it works fine from any user, need to check, and if answer is yes, remove user check.

aalexfvk:
> If I remember correct the reason was that when `ch-monitoring` starter from `root`, it creates some artifacts like log files with owner `root:root`. And later when it starts from user `monitor` it fails with access denied error. So `sudo /usr/bin/ch-monitoring ping` must stops with `Wrong current user: root`.

`sudo /usr/bin/ch-monitoring ping` should work because in this case an effective user is set here
https://github.com/yandex/ch-tools/blob/66778cc76e94c4a092724c3bfa412cb1344a4cd9/ch_tools/monrun_checks/main.py#L165-L173
We shouldn't see `Wrong current user: root` at all.

Moreover in the case I have fixed we see `Wrong current user: root` when effective user is `monitor` actually as I showed in the examples.

aalexfvk:
> PS. May be behavior was changed since that time and now it works fine from any user, need to check, and if answer is yes, remove user check.

The behaviour hasn't been changed. If we eliminate the code for user checking it still raises an exception after starting under root previosly.
I propose to fix the current approach because it leads to hanging to a couple of hours highstate now.

Alex-Burmak (approved):

PR #71, opened 2023-10-05
Mdb 25318 cleanup after replicated databases

MikhailBurdukov:
Logic is the same as table cleanup, but simpler: We are just deleting host from replicas and database node if it is empty, setting `is_lost` flag and  `queue` removing is not needed.

Also refactored: 2 `_find_paths` were replaced with 1 for `_find_tables` and `_remove_hosts_nodes`(Ones iterate through the tree and reuse the result)

### DIFF
--- a/tests/features/chadmin_zookeeper.feature
+++ b/tests/features/chadmin_zookeeper.feature
@@ -43,13 +43,29 @@ Feature: keeper-monitoring tool
     """
 
 
-  Scenario: Cleanup from Replicated database.
+  Scenario: Remove single host from Replicated database.
     # Imitate that we got nodes in zk from replicated database.
     When we execute chadmin create zk nodes on zookeeper01
     """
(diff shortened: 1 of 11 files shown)


tests/features/chadmin_zookeeper.feature
```diff
@@ -41,3 +41,31 @@ Feature: chadmin zookeeper commands.
     """
     /test/read_sli_part/shard1/replicas/host2.net
     """
+
+
+  Scenario: Remove single host from Replicated database.
+    # Imitate that we got nodes in zk from replicated database.
+    When we execute chadmin create zk nodes on zookeeper01
```
Alex-Burmak, line 48: It's rather out of scope of this PR, but it would be great to create real replicated tables and databases rather than mock them by manually creating znodes.

Alex-Burmak (approved):

PR #76, opened 2023-11-14
Use keeper hostname to match ssl cert

Nuzhnov:


myrrc (approved):
LGTM but let's forbid old TLS versions

PR #77, opened 2023-11-20
Config file support

Alex-Burmak:


myrrc (changes requested):
LGTM but maybe we should add a test loading a custom config and also one loading an invalid config file (rights issue, non-existent issue, bad contents)?

aalexfvk (approved):

myrrc:
Tests will be added in a separate PR

myrrc (approved):

PR #78, opened 2023-11-23
Add --keep-going option to "partition attach" and "partition detach" commands

Alex-Burmak:


ch_tools/chadmin/cli/partition_group.py
```diff
@@ -119,20 +120,38 @@ def list_partitions_command(ctx, **kwargs):
...
+            attach_partition(
+                ctx,
+                database=p["database"],
+                table=p["table"],
+                partition_id=p["partition_id"],
+                dry_run=dry_run,
+            )
+        except Exception as e:
+            if keep_going:
+                print(repr(e))
```
aalexfvk, line 152: Why not logging ?

Alex-Burmak: To not hide error from a user and output it to stdout.

Here's an example output: 

```
# chadmin partition attach --all -k
ALTER TABLE `db1`.`events` ATTACH PARTITION ID '20230306'

ClickhouseError("Code: 499. DB::Exception: The specified key does not exist.: while reading key: cloud_storage/cluster1001/shard1/r0000000000000000000000000000000000000000000011110011101111111001-file-ojvazctfenasiyjtsvzzsjmmubpdmgft, from bucket: cloud-storage-cluster1001. (S3_ERROR) (version 23.8.6.16 (official build))\n\nQuery: ALTER TABLE `db1`.`events` ATTACH PARTITION ID '20230306'",)

ALTER TABLE `db1`.`events` ATTACH PARTITION ID '20230326'
```

We can add logging in addition to outputting error to stdout. And error reporting to stdount not as good as it can be. But for now I implemented `--keep-going` option in consistent way with other commands that support it.

aalexfvk (approved):

PR #79, opened 2023-11-24
Use credentials from config in monitoring checks

Alex-Burmak:


ch_tools/common/clickhouse/client/clickhouse_client.py
```diff
@@ -136,13 +136,7 @@ def clickhouse_client(ctx):
...
         config = ctx.obj["config"]["clickhouse"]
-
-        user = config["user"]
-        password = config["password"]
-        if ctx.obj.get("monitoring", False) and config["monitoring_user"]:
-            user = config["monitoring_user"]
-            password = config["monitoring_password"]
-
+        user, password = clickhouse_credentials(ctx)
         ctx.obj["chcli"] = ClickhouseClient(
```
aalexfvk, line 140: Why do we maintain two `ClickhouseClient` classes ? This and from [monrun_checks](https://github.com/yandex/ch-tools/tree/4996b7a7d70fc451cbc2c2e5e0def7f3af59ef9b/ch_tools/monrun_checks/clickhouse_client.py).
Could we use only from `ctx.obj["chcli"]` ?

Alex-Burmak: Agree, it's redundant and confusing. It's better to leave only one client, but right now these two clients have no feature parity. The client from `monrun_checks` supports auto-detection of ports to use from config file, supports multiple ClickHouse protocols, but lack some other features.

In this PR I limited the changes to one small thing - use credentials from config file when connecting to ClickHouse in monitoring commands / monrun checks.

aalexfvk: Ok, I see. I'm going to create a task for merging them

aalexfvk (approved):

PR #89, opened 2024-01-09
Speed up cleanup-removed-hosts-metadata script

MikhailBurdukov:
Performance improvement for `cleanup-removed-hosts-metadata`. In this implementation we unite the nodes to delete in transactions to do single operation for batch of nodes.

Script for time comparison:
`cat populate.sh `
```
clickhouse client --query "DROP DATABASE IF EXISTS test_db"
clickhouse client --query "CREATE DATABASE test_db"

for i in $(seq 1 $1);
do
    
	
    clickhouse client --query "CREATE TABLE test_db.test$i (a int, b int) engine=ReplicatedMergeTree('/clickhouse/tables/{shard}/test_db.table$i', '{replica}') order by a partition by a"
    clickhouse client --query "SYSTEM STOP MERGES test_db.table$1"
    for j in $(seq 1 $2);
    do
      clickhouse client --query "INSERT INTO test_db.test1 SELECT rand()%10000, rand(1)%1000 FROM numbers(100)"
    done

done
sudo service clickhouse-server stop
```
Results:
cmd : `time chadmin zookeeper cleanup-removed-hosts-metadata  <host>`
| data | new |  old | 
|---|---|---|
|  ./populate.sh 10 20  |  0m15.114s  | 2m46.153s  | 
|  ./populate.sh 20 2  |  0m4.864s  |  0m51.093s  | 
|  ./populate.sh 2 40  |  0m8.673s  |  5m54.860s  | 
|  ./populate.sh 15 15  |  2m46.116s  |  8m42.093s  |

### DIFF
--- a/ch_tools/chadmin/internal/zookeeper.py
+++ b/ch_tools/chadmin/internal/zookeeper.py
@@ -171,23 +171,24 @@ def _delete_nodes_transaction(zk, to_delete_in_trasaction):
 
 def _remove_subpaths(paths):
     """
-    Removing from the list paths that are subpath of antoher.
+    Removing from the list paths that are subpath of another.
 
     Example:
     [/a, /a/b/c<-remove it]
     """
+    if not paths:
+        return
     # Sorting the list in the lexicographic order
     paths.sort()
     paths = [path.split("/") for path in paths]
-    normalized = [paths[0]]
-    # If path[i] have subnode path[j] then all paths from i to j will be subnode of i.
+    normalized_paths = [paths[0]]
+    # If path[i] has subnode path[j] then all paths from i to j will be subnode of i.
     for path in paths:
-        if (
-            len(normalized[-1]) > len(path)
-            or path[: len(normalized[-1])] != normalized[-1]
-        ):
-            normalized.append(path)
-    return ["/".join(normalized_list) for normalized_list in normalized]
+        last = normalized_paths[-1]
+        # Ignore the path if the last normalized one is its prefix
+        if len(last) > len(path) or path[: len(last)] != last:
+            normalized_paths.append(path)
+    return ["/".join(path) for path in normalized_paths]
 
 
 def _delete_recursive(zk, paths):
--- a/ch_tools/chadmin/internal/zookeeper.py
+++ b/ch_tools/chadmin/internal/zookeeper.py
@@ -194,7 +194,7 @@ def _delete_recursive(zk, paths):
                 normalized.append(path)
         return normalized
 
-    if not len(paths):
+    if len(paths) == 0:
         return
     print("Node to recursive delete", paths)
     paths = remove_subpaths(paths)
--- a/ch_tools/chadmin/internal/zookeeper.py
+++ b/ch_tools/chadmin/internal/zookeeper.py
@@ -2,10 +2,11 @@ import logging
 import os
 import re
 from contextlib import contextmanager
+from math import sqrt
 from queue import Queue
 
 from kazoo.client import KazooClient
-from kazoo.exceptions import NoNodeError, NotEmptyError
+from kazoo.exceptions import NoNodeError
 
 from ch_tools.chadmin.cli import get_clickhouse_config, get_macros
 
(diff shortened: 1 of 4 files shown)


ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -148,22 +143,63 @@ def _find_paths(zk, root_path, included_paths_regexp, excluded_paths_regexp=None
...
+    Kazoo already has the ability to recursively delete nodes, but the implementation is quite naive
+    and has poor performance with a large number of nodes being deleted.
+
+    In this implementation we unite the nodes to delete in transactions to do single operation for batch of nodes.
+    To delete in correct order first of all we perform topological sort using bfs approach.
+    """
+
+    print("Node to recursive delete", paths)
+    nodes_to_delete = []
+    queue: Queue = Queue()
```
aalexfvk, line 178: Actually, [`queue.Queue`](https://docs.python.org/3/library/queue.html) is intended for communication in threaded apps and has corresponding overhead. It is much more preferable to use [`collections.deque`](https://docs.python.org/3/library/collections.html?highlight=collections#collections.deque) as FIFO data structure.

It can be initialized as `d = deque(list_obj)`

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -148,22 +143,63 @@ def _find_paths(zk, root_path, included_paths_regexp, excluded_paths_regexp=None
...
+    for path in paths:
+        queue.put(path)
+        nodes_to_delete.append(path)
+
+    while not queue.empty():
+        path = queue.get()
+        for child_node in _get_children(zk, path):
+            full_child_path = os.path.join(path, child_node)
+            queue.put(full_child_path)
+            nodes_to_delete.append(full_child_path)
```
aalexfvk, line 188: Is it correct to put children nodes to  `nodes_to_delete` here ?
Wouldn't that be more correct ?
```
    while not queue.empty():
        path = queue.get()
        nodes_to_delete.append(path)
        for child_node in _get_children(zk, path):
            queue.put(os.path.join(path, child_node))            
```

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -148,22 +143,63 @@ def _find_paths(zk, root_path, included_paths_regexp, excluded_paths_regexp=None
...
             #  Someone deleted node before us. Do nothing.
             print("Node {node} is already absent, skipped".format(node=node))
-        except NotEmptyError:
-            #  Someone created child node while deleting.
-            #  I'm not sure that we can get this exception with recursive=True.
-            #  Do nothing.
-            print("Node {node} is not empty, skipped".format(node=node))
+
+
+def _delete_recursive(zk, paths):
```
aalexfvk, line 167: What if `paths` contains parent children node paths ?

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -148,22 +143,63 @@ def _find_paths(zk, root_path, included_paths_regexp, excluded_paths_regexp=None
...
+        nodes_to_delete.append(path)
+
+    while not queue.empty():
+        path = queue.get()
+        for child_node in _get_children(zk, path):
+            full_child_path = os.path.join(path, child_node)
+            queue.put(full_child_path)
+            nodes_to_delete.append(full_child_path)
+
+    operations_in_transaction = max(100, int(sqrt(len(nodes_to_delete))))
```
aalexfvk, line 190: When is `sqrt` value useful ?

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -148,22 +143,63 @@ def _find_paths(zk, root_path, included_paths_regexp, excluded_paths_regexp=None
...
-    def _try_delete_zk_node(zk, node):
+    delete_transaction = zk.transaction()
+    for node in to_delete_in_trasaction:
+        delete_transaction.delete(node)
+    result = delete_transaction.commit()
+
+    if result.count(True) == len(result):
+        # Transaction completed successfully, exit.
+        return
+    for node in to_delete_in_trasaction:
```
aalexfvk, line 159: Let's add log message about the fallback to sequential deleting

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -143,27 +138,88 @@ def _find_paths(zk, root_path, included_paths_regexp, excluded_paths_regexp=None
...
+        return normalized
+
+    if not len(paths):
+        return
+    print("Node to recursive delete", paths)
+    paths = remove_subpaths(paths)
+    nodes_to_delete = []
+    queue = deque(paths)
+
+    while len(queue):
```
aalexfvk, line 204: `while queue` is sufficient

aalexfvk (approved):

PR #91, opened 2024-01-23
Wait replication sync command

kirillgarbar:
Wait replication sync script is moved to ch-monitoring since it is basically a call to ch-monitoring command inside.
It might be worth moving it to chadmin.

I also not sure how we should implement a call of a click command inside another command, the most direct way to do it is to call a command with `ctx.invoke()` or `ctx.forward()`, but the [docs](https://click.palletsprojects.com/en/8.0.x/advanced/#invoking-other-commands) suggest it is not a recommended way to do this. That's why I moved `replication-lag` command to a serapate method for a while.


ch_tools/monrun_checks/ch_replication_sync.py
```diff
@@ -0,0 +1,42 @@
...
+    "-p",
+    "--pause",
+    type=int,
+    default=30,
+    help="Pause between request in seconds, default is 30 seconds.",
+)
+@option(
+    "-t",
+    "--timeout",
+    type=int,
```
Alex-Burmak, line 27: Suggest to use `TimeSpanParamType` 

Usage example:
https://github.com/yandex/ch-tools/blob/baaf84329cb7f524eef08e8256c6537879934d81/ch_tools/chadmin/cli/object_storage_group.py#L123-L131

ch_tools/monrun_checks/ch_replication_sync.py
```diff
@@ -0,0 +1,42 @@
...
+    default=3 * 24 * 60 * 60,
+    help="Max amount of time to wait, in seconds. Default is 30 days.",
```
Alex-Burmak, line 29: Default is 3 days, right? Not 30.

And default values should be automatically added to help messages. So it's not required to write them manually.

ch_tools/monrun_checks/ch_replication_sync.py
```diff
@@ -0,0 +1,42 @@
...
+)
+@pass_context
+def wait_replication_sync_command(ctx, status, pause, timeout):
```
Alex-Burmak, line 32: Suggest to move the command to `chadmin` and `estimate_replication_lag` function from `ch_replication_lag.py` somewhere to https://github.com/yandex/ch-tools/blob/master/ch_tools/common

ch_tools/monrun_checks/ch_replication_lag.py
```diff
@@ -56,6 +56,10 @@
 )
 @click.pass_context
 def replication_lag_command(ctx, xcrit, crit, warn, mwarn, mcrit, verbose):
+    return estimate_replication_lag(ctx, xcrit, crit, warn, mwarn, mcrit, verbose)
+
+
+def estimate_replication_lag(ctx, xcrit=3600, crit=6000, warn=300, mwarn=50.0, mcrit=90.0, verbose=0):
```
aalexfvk, line 62: Lets move hardcoded values to constants in top of file. They might be mismatched. And it seems already are (`crit = 600` above).

ch_tools/monrun_checks/ch_replication_sync.py
```diff
@@ -0,0 +1,42 @@
...
+    """Wait for ClickHouse server to sync replication with other replicas."""
+
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        res = estimate_replication_lag(ctx)
```
aalexfvk, line 37: Are default arguments for `estimate_replication_lag` suitable for this command ?

ch_tools/monrun_checks/ch_replication_sync.py
```diff
@@ -0,0 +1,42 @@
...
+        if res.code <= status:
+            return Result(code=0) 
+        time.sleep(pause)
+
+    return Result(code=2, message=f"ClickHouse can\'t sync replica for {timeout} seconds")
```
aalexfvk, line 42: nit: Escaping for `'` is not needed in the `""` string

ch_tools/monrun_checks/ch_replication_sync.py
kirillgarbar: In the original script replication-lag command is used with the default arguments. We may add some more arguments to the wait command and pass them here but I don't like the idea to bound them to the command implementation this way

tests/features/monrun.feature
```diff
@@ -139,6 +139,23 @@ Feature: ch-monitoring tool
...
+    """
+    And we sleep for 5 seconds
+    And we execute command on clickhouse01
+    """
+    ch-monitoring replication-lag -w 4
+    """
+    Then we get response contains
+    """
+    1;
+    """
```
kirillgarbar, line 158: I was thinking about the same test that detects replication lag for `replication-sync` command, but default timeout for `replication-lag` is 300 seconds. This would require to sleep for 300 seconds and will slow tests a lot.
Maybe we should pass args for `replication-lag` to `replication-sync` as discussed above to avoid this

kirillgarbar:
Also I discovered that running a `chadmin` in tests causes import exception. The reason is `urllib3` version > 2 is installed in clickhouse container despite the fact that it is locked to < 1.27 in lock file.
To fix this locally I had to add `RUN pip install --force-reinstall -v "urllib3==1.26.5"` to clickhouse docker file.

tests/features/monrun.feature
aalexfvk: I think it makes sense

aalexfvk:
> Also I discovered that running a `chadmin` in tests causes import exception. The [reason](https://github.com/psf/requests/issues/6443) is `urllib3` version > 2 is installed in clickhouse container despite the fact that it is locked to < 1.27 in lock file. To fix this locally I had to add `RUN pip install --force-reinstall -v "urllib3==1.26.5"` to clickhouse docker file.

At first sight, we can pin `requests = "<2.30.0"` in pyproject.toml as a workaround

PR #97, opened 2024-01-30
Wait replication sync using SYNC REPLICA

kirillgarbar:
Not sure what default timeout value for SYNC REPLICA should be.
Command throws on SYNC REPLICA timeout. I thought we can just retry sync until total_timeout is reached, but I think current implementation allows us to know that something is not as expected much earlier.
Replication lag check is not removed in case something really weird happens.
Also I just print() on timeout because throwing exceptions result in some ugly traceback and useless message being printed, maybe there is a better way to do it.

### DIFF
--- a/ch_tools/chadmin/cli/wait_group.py
+++ b/ch_tools/chadmin/cli/wait_group.py
@@ -9,6 +9,7 @@ from ch_tools.chadmin.internal.utils import execute_query
 from ch_tools.common.cli.parameters import TimeSpanParamType
 from ch_tools.common.commands.replication_lag import estimate_replication_lag
 from ch_tools.common.utils import execute
+from requests.exceptions import ReadTimeout
 
 BASE_TIMEOUT = 600
 LOCAL_PART_LOAD_SPEED = 10  # in data parts per second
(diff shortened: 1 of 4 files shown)


ch_tools/chadmin/cli/wait_group.py
```diff
@@ -76,29 +84,55 @@ def wait_group():
...
-    deadline = time.time() + timeout.total_seconds()
+    """Wait for ClickHouse server to sync replication with other replicas."""
+
+    start_time = time.time()
+    deadline = start_time + total_timeout.total_seconds()
+
+    # Get replicated tables
+    tables = execute_query(
+        ctx,
+        "SELECT name, database FROM system.tables WHERE engine = 'ReplicatedMergeTree'",
```
aalexfvk, line 108: What about other replicated engines like `ReplicatedCollapsingMergeTree` ?

aalexfvk:
LGTM, it would be great if someone else take a look

ch_tools/chadmin/cli/wait_group.py
```diff
@@ -76,29 +85,62 @@ def wait_group():
...
+
+        try:
+            execute_query(
+                ctx,
+                f"SYSTEM SYNC REPLICA {full_name}",
+                format_=None,
+                timeout=timeout,
+                settings={"receive_timeout": timeout},
+            )
+        except ReadTimeout:
```
aalexfvk, line 127: Could you cover both of these `except` branches in tests ?

kirillgarbar: Currently we can't control which exception will be raised if equal timeout is set. I thought we can disable something on clickhouse side to wait for ReadTimeout, but it will probably raise other errors like connection timeout, etc.

Alex-Burmak (approved):

PR #100, opened 2024-02-07
Change ddl tasks  state to finished for removed hosts.

MikhailBurdukov:


ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -340,10 +354,49 @@ def databases_cleanups(zk, zk_root_path):
...
         _remove_if_no_hosts_in_replicas(zk, databases_paths)
 
+    def get_host_mention_in_task(zk, host, ddl_task_path):
+        """
+        Predicate for indicating that host must execute this task.
+
+        If there is a host mention in the ddl, then returns the escaped hostname with port.
+        """
+        try:
+            for line in zk.get(ddl_task_path)[0].decode().strip().splitlines():
```
aalexfvk, line 364: Using https://github.com/yandex/ch-tools/blob/0ba8de49db877e2f087ff7ddec69f2c4fe9fbcd7/ch_tools/chadmin/internal/zookeeper.py#L15-L19 will make the code more readable imho

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -340,10 +354,49 @@ def databases_cleanups(zk, zk_root_path):
...
+                if "hosts: " in line:
```
aalexfvk, line 365: Please write in docstring what structure of node value we expect

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -340,10 +354,49 @@ def databases_cleanups(zk, zk_root_path):
...
+                    pos = line.find(escaped_host)
+                    # Not found
+                    if pos == -1:
+                        return None
+
+                    return line[pos : line.find("'", pos)]
+        except NoNodeError:
+            pass
+
+    def propagate_ddl_query(zk):
```
aalexfvk, line 376: Non-obvious naming. How about `finish_ddl_query` or `mark_finished_ddl_query` ?

tests/features/chadmin_zookeeper.feature
```diff
@@ -88,3 +88,42 @@ Feature: chadmin zookeeper commands.
     And we delete zookeepers nodes /test/a,/test/a/b on zookeeper01
     And we delete zookeepers nodes /test/c/d,/test/c on zookeeper01
     Then the list of children on zookeeper01 for zk node /test are empty
+
+
+  Scenario: Set finished to ddl with removed host.
+    # Create the fake cluster with removed host. 
+    When we execute command on clickhouse02
+    """
+    echo -e "
```
aalexfvk, line 97: Let's write a [step](https://github.com/yandex/ch-tools/blob/0ba8de49db877e2f087ff7ddec69f2c4fe9fbcd7/tests/steps/clickhouse.py) for this action, e.g. `When we put the  clickhouse config to path `/a/b/c.xml` with/without restarting`. And leave here only config content and hide shell specific things in the step.

ch_tools/chadmin/internal/zookeeper.py
MikhailBurdukov: Think that it is hard to reuse this method, because we already have formatted paths and opened zk connection

aalexfvk (approved):

PR #106, opened 2024-02-25
Common clickhouse client

kirillgarbar:
Clickhouse clients in tests and monitoring replaced with common client and wrappers around it.
Port resolution code with http and tcp execution methods is moved to common client.


ch_tools/monrun_checks/utils.py
```diff
@@ -0,0 +1,66 @@
+from ch_tools.common.clickhouse.client.clickhouse_client import (
+    ClickhousePort,
+    clickhouse_client,
+)
+
+
+def execute_query_client(
```
aalexfvk, line 7: How about moving this function to `ClickhouseClient` as member function `query_json_data()`
We can use it also in `replication_lag.py`

ch_tools/monrun_checks/utils.py
```diff
@@ -0,0 +1,66 @@
...
+        format_=format_,
+        echo=echo,
+        dry_run=dry_run,
+        stream=stream,
+        settings=settings,
+        port=port,
+    )["data"]
+
+
+def execute_query(
```
aalexfvk, line 39: What benefit of this proxy function ? Could we use `clickhouse_client()` instead ?

aalexfvk (approved):

PR #113, opened 2024-03-26
Return warning on user's fault backup fail reasons

kirillgarbar:
Backups that failed with user's fault now will cause `WARNING` status instead of `CRIT`

### DIFF
--- a/ch_tools/monrun_checks/ch_backup.py
+++ b/ch_tools/monrun_checks/ch_backup.py
@@ -31,7 +31,7 @@ FAILED_PARTS_THRESHOLD = 10
     "failed_backup_count_crit_threshold",
     default=3,
     help="Critical threshold on the number of failed backups. If last backups failed and its count equals or greater "
-    "than the specified threshold, a crit will be reported.",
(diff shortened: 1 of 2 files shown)


Alex-Burmak:
Relates to https://github.com/yandex/ch-backup/pull/114

ch_tools/monrun_checks/ch_backup.py
```diff
@@ -112,23 +112,30 @@ def _check_last_backup_not_failed(backups: List[Dict], crit_threshold: int) -> R
...
 
-    if counter == 1:
+    if total_count == 1:
         message = "Last backup failed"
     else:
-        message = f"Last {counter} backups failed"
+        message = f"Last {total_count} backups failed"
 
     status = CRIT if crit_threshold and counter >= crit_threshold else WARNING
     return Result(status, message)
```
Alex-Burmak, line 141: If we suppress crit due to user fault errors, it makes sense to add corresponding note to the result message. Something like `... (suppressed due to user fault errors)". There are similar functionality in `_suppress_if_required` function.

Alex-Burmak (approved):

PR #115, opened 2024-04-03
Fix macros substition

Alex-Burmak:


ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -100,7 +100,7 @@ def _format_path(ctx, path):
     no_ch_config = args.get("no_ch_config", False)
     if no_ch_config:
         return path
-    return path.format_map(get_macros(ctx))
```
Alex-Burmak, line 103: `format_map` doesn't work in corner cases, see https://github.com/yandex/ch-backup/pull/116

MikhailBurdukov (approved):

PR #116, opened 2024-04-04
Enable command for sigle table cleanup in zk and refactor tests

MikhailBurdukov:


ch_tools/chadmin/cli/zookeeper_group.py
```diff
@@ -308,4 +308,23 @@ def clickhouse_hosts_command(ctx, fqdn):
     # We can't get the ddl queue path from clickhouse config,
     # because in some cases we are changing this path while performing cluster resetup.
     config = load_config()
```
Alex-Burmak, line 310: Unrelated, but loading config is redundant here. It's already loaded and available in the context. It can be retrieved with `ctx.obj["config"]`.

Alex-Burmak (approved):

aalexfvk (approved):

PR #117, opened 2024-04-05
Add set-flag command

kirillgarbar:
New command is added to the table command group.
It sets flag inside table's data directory.

### DIFF
--- a/ch_tools/chadmin/cli/table_group.py
+++ b/ch_tools/chadmin/cli/table_group.py
@@ -3,15 +3,11 @@ import os
 from click import Choice, Context, argument, group, option, pass_context
 
 from ch_tools.chadmin.cli import get_cluster_name
-from ch_tools.chadmin.internal.system import match_ch_version
 from ch_tools.chadmin.internal.table import (
-    CONVERT_TO_REPLICATED_FLAG,
     attach_table,
     delete_table,
     detach_table,
     get_table,
-    get_tables_dict,
-    get_tables_to_convert,
     list_tables,
     materialize_ttl,
 )
@@ -376,7 +372,12 @@ def materialize_ttl_command(ctx, dry_run, all_, database, table, exclude_table):
 @option(
     "--database",
     type=StringParamType(),
-    help="Filter tables to set the flag by the specified database name.",
+    help="Filter in tables to set the flag by the specified database name.",
+)
+@option(
+    "--exclude-database",
+    type=StringParamType(),
+    help="Filter out tables to set the flag by the specified database name.",
 )
 @option(
     "--table",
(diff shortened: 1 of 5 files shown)


ch_tools/chadmin/cli/table_group.py
```diff
@@ -363,3 +369,59 @@ def materialize_ttl_command(ctx, dry_run, all_, database, table, exclude_table):
...
+    Create a flag with the specified name inside the data directory of the table.
+    """
+    if database is None:
+        return
+
+    tables = (
+        get_tables_to_convert(ctx, database, table, exclude_table)
+        if flag == CONVERT_TO_REPLICATED_FLAG
+        else get_tables_dict(ctx, database, table, exclude_table)
+    )
```
kirillgarbar, line 417: We can also make this command more generic and make user's responsibility to filter tables with not MergeTree engine, but currently there are no other table flags in clickhouse other than convert_to_replicated.

kirillgarbar:
I am also not sure if we should enable this feature until convertation fix is released, currently it may not work with tables that do not use default storage policy

ch_tools/chadmin/cli/table_group.py
```diff
@@ -363,3 +370,67 @@ def materialize_ttl_command(ctx, dry_run, all_, database, table, exclude_table):
...
+    exclude_table: str,
+    flag: str,
+    verbose: bool,
+) -> None:
+    """
+    Create a flag with the specified name inside the data directory of the table.
+    """
+    if database is None:
+        print("Database is not specified")
+        return
```
Alex-Burmak, line 413: Can be replaced with `required=True` in `@option` definition.

But I suggest to do make the check less strict: 
```
    if not any((all_, database, table)):
        ctx.fail(
            "At least one of --all, --database, --table options must be specified."
        )
```

ch_tools/chadmin/cli/table_group.py
```diff
@@ -363,3 +370,67 @@ def materialize_ttl_command(ctx, dry_run, all_, database, table, exclude_table):
...
+@option(
+    "--table",
+    type=StringParamType(),
+    help="Filter in tables by the specified table name.",
+)
+@option(
+    "--exclude-table",
+    type=StringParamType(),
+    help="Filter out tables by the specified table name.",
+)
```
Alex-Burmak, line 390: Suggest to also add options `--exclude-database` and `--all`.

ch_tools/chadmin/internal/table.py
```diff
@@ -183,3 +189,36 @@ def materialize_ttl(ctx, database, table, echo=False, dry_run=False):
...
+    )
+    return [
+        {
+            "database": item["database"],
+            "table": item["table"],
+            "data_paths": item["data_paths"],
+            "engine": item["engine"],
+        }
+        for item in tables
+    ]
```
Alex-Burmak, line 224: Looks like a redundant logic. Why do not return `list_tables` result as is?

ch_tools/chadmin/cli/table_group.py
Alex-Burmak: I like this idea. Let's do it.

And in this case I'd suggest to replace `--flag` `@option` with mandatory `@argument`.

Then setting the flag `convert_to_replicated` for all tables will look something like this: 
```
chadmin table set-flag --all --exclude-database system,_system --engine %MergeTree --exclude-engine Replicated% convert_to_replicated
```

Alex-Burmak (changes requested):

Alex-Burmak:
> I am also not sure if we should enable this feature until convertation fix is released, currently it may not work with tables that do not use default storage policy

I think it's OK to not wait for the fix.

Alex-Burmak (approved):

PR #118, opened 2024-04-08
Add backoff for s3 based queries

MikhailBurdukov:


kirillgarbar (approved):

ch_tools/common/clickhouse/client/clickhouse_client.py
```diff
@@ -230,17 +238,36 @@ def query(
...
+                        format_,
+                        post_data,
+                        timeout,
+                        stream,
+                        per_query_settings,
+                        found_port,
+                    )
+
+                return self._execute_tcp(query, format_, found_port)
+            except ClickhouseError as e:
```
aalexfvk, line 256: Handling of retries here somewhat clutters this function. I'd suggest moving retry logic in a separate class or function and pass it  `self._execute_http()` as a callable.

aalexfvk:
Does this retries produce duplicates in CH ?

MikhailBurdukov:
> Does this retries produce duplicates in CH ?

Do you mean that we can have duplicates in the table after repeated insertion attempt with multiple values?

PR #120, opened 2024-04-09
Use local storage policy for object storage cleanup

MikhailBurdukov:


ch_tools/chadmin/cli/object_storage_group.py
```diff
@@ -111,6 +111,12 @@ def object_storage_group(ctx: Context, disk_name: str) -> None:
     is_flag=True,
     help=("Use saved object list without traversing object storage again."),
 )
+@option(
+    "--storage-policy",
+    "storage_policy",
+    default="local",
+    help=("Internal table storage policy.."),
+)
```
Alex-Burmak, line 119: Suggest to add a config file option instead. It would be more convenient to use / integrate in YC. Adding command-line option in additional to config file option is optional.

Also suggest to use `default` storage policy by default. `local` is not a standard storage policy and can be missing in ClickHouse clusters outside YC. Another option is to not specify any storage policy by default.

Alex-Burmak (changes requested):

Alex-Burmak (approved):

PR #121, opened 2024-04-09
Specify disk for clean orphaned data and some pretty printing

MikhailBurdukov:


ch_tools/chadmin/cli/data_store_group.py
```diff
@@ -30,40 +33,89 @@ def data_store_group():
...
-
-        process_path(path, prefix, column, remove)
-
-
-def process_path(path: str, prefix: str, column: str, remove: bool) -> None:
-    print(f"Processing path {path} with prefix {prefix}:")
+@option(
+    "--store-path",
+    "store_path",
+    default=CLICKHOUSE_PATH + "/store",
```
aalexfvk, line 39: What is the point to remove `CLICKHOUSE_STORE_PATH` ? It seems useful constant

ch_tools/chadmin/cli/data_store_group.py
```diff
@@ -30,40 +33,89 @@ def data_store_group():
...
+    help="Set the store subdirectory path,  `/var/lib/clickhouse/store` by default.",
```
aalexfvk, line 40: I think we can remove `by default` clause here. Because we already show default values by this setting https://github.com/yandex/ch-tools/blob/6b9c5d2c441f61a7c01c9ab82768f55bec2cb6b2/ch_tools/common/cli/context_settings.py#L13

ch_tools/chadmin/cli/data_store_group.py
```diff
@@ -30,40 +33,89 @@ def data_store_group():
...
+    ctx, column, remove, store_path, show_directory_size, show_only_orphaned_metadata
+):
+    results = []
+    for prefix in os.listdir(store_path):
+        path = store_path + "/" + prefix
+        path_result = process_path(path, prefix, column, remove, show_directory_size)
+        if show_only_orphaned_metadata and path_result["status"] != "not_used":
+            continue
+        if not show_directory_size:
+            del path_result["size"]
```
aalexfvk, line 66: When we don't want to see size in output ?

MikhailBurdukov: Have removed the option. Seems like we always want to see sizes.

aalexfvk (approved):

PR #136, opened 2024-04-15
Improved error handling, dependency checks, and code readability

Adriel007:
# Improvements to the Bash Script

In this document, we describe the changes made to the provided Bash script to add error handling and suggest a security enhancement.

## Added Error Handling

- A `error_exit()` function has been added to display error messages and exit the script when necessary.
- Additional checks have been added to ensure the script behaves correctly in case of issues.

## Security Enhancement Suggestion

It would be beneficial to enhance the security of the signing key. Currently, the script handles signing keys in plain text, which can be risky, especially in shared environments. It is recommended to consider using recommended security practices, such as storing keys in a key vault or using a secure key management mechanism.

### DIFF
--- a/build_deb.sh
+++ b/build_deb.sh
@@ -2,59 +2,42 @@
 
 set -e
 
-# Function to check the availability of a command
-check_command() {
-    if ! command -v "$1" &> /dev/null; then
-        echo "Error: '$1' command not found. Please install it before running this script." >&2
-        exit 1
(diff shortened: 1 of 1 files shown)


aalexfvk:
I agree with you. We have only used package build in our private environment so far. But security enhancements will be considered soon.

aalexfvk (approved):

PR #141, opened 2024-04-17
Logging with loguru

kirillgarbar:
Common logging module for all ch-tools components, implemented similar to ch-backup.
For some reason loguru is using only kwargs for formatting, so I replaced %-formatting with f-strings.
Prints will be replaces in the next PR, because this is already quite large.


ch_tools/common/logging.py
```diff
@@ -0,0 +1,182 @@
...
+
+import inspect
+import logging
+import sys
+from typing import Any, Optional
+
+from loguru import logger
+
+# Used to store this.module variable
+this = sys.modules[__name__]
```
MikhailBurdukov, line 13: Do we really need to put logger name in `sys.modules`? I think we can replace it with  dictionary to keep name and option parameters. besides, we'll get rid of replacing parameters in the format string: https://github.com/kirillgarbar/ch-tools/blob/4df613ecf6a1c8d7cf683205945cbacfa80ee7d6/ch_tools/common/logging.py#L80

ch_tools/common/logging.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+    with_exception = kwargs.get("exc_info", False)
+    getLogger(this["module"]).opt(exception=with_exception).debug(msg, *args, **kwargs)
+
+
+def log_status(status, msg):
+    """
+    Log a message with by status code.
+    """
+    if status == 0:
```
Alex-Burmak, line 162: Suggest to use constants OK, WARNING and ERROR from ch_tools/common/result.py instead of magic constants.

See also https://github.com/yandex/ch-tools/pull/143 This PR replaces 0, 1, 2 with constants in almost all places in the repo.

ch_tools/monrun_checks_keeper/main.py
```diff
@@ -71,9 +66,7 @@ def wrapper(ctx, *a, **kw):
                 status.set_code(1)
 
             log_message = f"Completed with {status.code};{status.message}"
-            logging.log(
-                logging.DEBUG if status.code == 0 else logging.ERROR, log_message
-            )
+            logging.log_status(status.code if status.code <= 1 else 2, log_message)
```
Alex-Burmak, line 69: Suggest to replace with
```
logging.log_status(status.code, log_message)
```

The old logic looks wrong / doesn't handle warnings correctly.

tests/modules/s3.py
```diff
@@ -45,7 +47,7 @@ def __init__(self, context: ContextT, bucket: Optional[str] = None) -> None:
         self._s3_bucket_name = bucket if bucket else config["bucket"]
 
         for module_logger in ("boto3", "botocore", "s3transfer", "urllib3"):
-            logging.getLogger(module_logger).setLevel(logging.CRITICAL)
+            logging.set_module_log_level(module_logger, logging.logging.CRITICAL)
```
Alex-Burmak, line 50: `logging.logging.CRITICAL` looks strange (double logging). Suggest to make CRITICAL available directly in `ch_tools.common.logging` namespace that it can be used as `logging.CRITICAL`.

The same is applicable for other log level constants and `getHandler` function.

ch_tools/common/logging.py
```diff
@@ -0,0 +1,189 @@
...
+Logging module.
+"""
+
+import inspect
+import logging
+from typing import Any, Dict, Optional
+
+from loguru import logger
+
+this: Dict[str, Any] = {}
```
MikhailBurdukov, line 11: 1. Les's keep in this dict two state: the name and extra options. In that case we can pass directly `this[opt]`  to logger:
```
getLogger(this["module"]).opt(this['opt'])...
```

2. Let's rename like `current_logger' or 'logger_config'.

ch_tools/common/logging.py
```diff
@@ -0,0 +1,189 @@
...
+        while frame and (depth == 0 or frame.f_code.co_filename == logging.__file__):
+            frame = frame.f_back
+            depth += 1
+
+        logger.bind(logger_name=record.name).opt(
+            depth=depth, exception=record.exc_info
+        ).log(level, record.getMessage())
+
+
+def configure(config_loguru: dict, module: str, cmd_name: Optional[str] = None) -> None:
```
MikhailBurdukov, line 68:
```suggestion
def configure(config_loguru: dict, module: str, extra_opt: dict: ) -> None:
```
IMO it is not necessary to use `cmd name`  as a parameter. Lets make it scalable.

MikhailBurdukov (approved):

ch_tools/common/logging.py
kirillgarbar: There are two kinds of parameters which we can add to logger. `cmd_name` is used in `bind()` so it will be added to `extra` dictionary and can be used in formatting. There are also `opt()` args, such as `exception`, there are no much usage for others, but I added both `extra` and `opt` arguments for `configure` just in case.

Alex-Burmak (approved):

ch_tools/common/logging.py
```diff
@@ -0,0 +1,213 @@
...
+        getLogger(logger_config["module"]).opt(**opt).critical(msg, *args, **kwargs)
+    elif level == ERROR:
+        getLogger(logger_config["module"]).opt(**opt).error(msg, *args, **kwargs)
+    elif level == WARNING:
+        getLogger(logger_config["module"]).opt(**opt).warning(msg, *args, **kwargs)
+    elif level == INFO:
+        getLogger(logger_config["module"]).opt(**opt).info(msg, *args, **kwargs)
+    elif level == DEBUG:
+        getLogger(logger_config["module"]).opt(**opt).debug(msg, *args, **kwargs)
+
```
MikhailBurdukov, line 138:
```suggestion
    getLogger(logger_config["module"]).opt(**opt).log(level, msg, *args, **kwargs)
```

ch_tools/common/logging.py
```diff
@@ -0,0 +1,213 @@
...
+Logging module.
+"""
+
+import inspect
+import logging
+from logging import (  # noqa # pylint:disable=unused-import
+    CRITICAL,
+    DEBUG,
+    ERROR,
+    FATAL,
```
MikhailBurdukov, line 11: Unused

ch_tools/common/logging.py
```diff
@@ -0,0 +1,213 @@
...
+    INFO,
+    WARN,
```
MikhailBurdukov, line 13: Unused

ch_tools/common/logging.py
```diff
@@ -0,0 +1,213 @@
...
+        logger.bind(logger_name=record.name).opt(
+            depth=depth, exception=record.exc_info
+        ).log(level, record.getMessage())
+
+
+def configure(
+    config_loguru: dict,
+    module: str,
+    extra: Optional[dict] = None,
+    opt: Optional[dict] = None,
```
MikhailBurdukov, line 83: Never used

ch_tools/common/logging.py
```diff
@@ -0,0 +1,213 @@
...
+
+    logger.configure(handlers=loguru_handlers, activation=[("", True)])
+
+    # Configure logging.
+    logging.basicConfig(handlers=[InterceptHandler()], level=0)
+
+
+def _log(level, msg, *args, **kwargs):
+    # Make a copy so we can rewrite exception option
+    opt = dict(logger_config["opt"])
```
MikhailBurdukov, line 124: afaik the opt is used only for single logging call. So it make sense to remove it from the global state

ch_tools/common/logging.py
Alex-Burmak: It's for addressing https://github.com/yandex/ch-tools/pull/141#discussion_r1570180044

ch_tools/common/logging.py
kirillgarbar: What did you mean by removing it from the global state? `logger_config["module"]` is also used just for logging call, passing these values from configure to a logging call is a reason why we store them in a dict.

MikhailBurdukov: I mean  remove  `opt` from  `logger_config`.  Because it never used and also everything that we can store in it, we should pass thought `logger_config['extra']`

MikhailBurdukov (approved):

PR #143, opened 2024-04-17
Output escaping rules

Alex-Burmak:
The changes make escaping rules tunable through config file.

Default set of escaping rules is minimal and doesn't include escaping rules for Prometheus-compatibility to better readability. 

_Without changes_
```
# ch-monitoring system-queues
2;test db.action: parts to check 34 > 20 (crit); test db.action: queue size 961 > 20 (crit); test db.action: merges in queue 17 > 10 (warn);
```

_With changes_
```
# ch-monitoring system-queues
2;test_db.action: parts_to_check 34 > 20 (crit); test_db.action: queue_size 961 > 20 (crit); test_db.action: merges_in_queue 17 > 10 (warn);
```

Now the output contains correct database and table names.

### DIFF
--- a/ch_tools/common/config.py
+++ b/ch_tools/common/config.py
@@ -46,6 +46,16 @@ DEFAULT_CONFIG = {
             },
         },
     },
+    "monitoring": {
+        "output": {
+            "escaping_rules": [
+                {
+                    "pattern": "\n",
+                    "replacement": "",
+                },
+            ],
+        },
+    },
     "ch-monitoring": {
         "log-errors": {
             "crit": 60,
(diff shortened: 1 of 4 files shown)


ch_tools/common/config.py
```diff
@@ -46,6 +46,16 @@
             },
         },
     },
+    "monitoring": {
+        "output": {
+            "escaping_rules": [
+                {
+                    "pattern": "\n",
+                    "replacement": "",
```
ianton-ru, line 54: Is any sample with EOL in output text? May be better make replacement with some delimiter (space) to change
```
foo
bar
```
on
```
foo bar
```
not
```
foobar
```
?

Alex-Burmak: Good idea! I've updated the replacement string to `" "`.

ianton-ru (approved):

PR #122, opened 2024-04-11
Support overriding thresholds through config

Alex-Burmak:

PR #151, opened 2024-04-20
Support of IPv6-only networks in DNS check

Alex-Burmak:


ch_tools/monrun_checks/dns.py
```diff
@@ -29,30 +29,35 @@ def __init__(self, fqdn: str, private: bool, strict: bool):
         self.strict = strict
 
 
-@click.command("ext_ip_dns")
```
Alex-Burmak, line 32: Renamed the command from `ext_ip_dns` to `dns` cause:
- `ext`prefix is incorrect as the command checks not only external IPs;
- The command name doesn't conform naming style of other commands - it's used `_` instead of `-` as a word separator.

ch_tools/monrun_checks/dns.py
```diff
@@ -29,30 +29,35 @@ def __init__(self, fqdn: str, private: bool, strict: bool):
...
+@click.command("dns")
 @click.option("--cluster", "cluster", is_flag=True, help="Check cluster DNS records")
 @click.option("--private", "private", is_flag=True, help="Check private DNS records")
+@click.option("--ipv4", "ipv4", is_flag=True, help="Check A DNS records")
```
Alex-Burmak, line 35: Enabled by default in `config.py`.

ianton-ru: Why not with `default=True` here?

Alex-Burmak: Just to not duplicate defaults from `config.py`.

After #122, if an option is controlled through both config and command-line argument it's possible to specify its default value in only one place - in `config.py`.

ianton-ru (approved):

PR #152, opened 2024-04-20
Update ruff linter version to 0.4.1

Alex-Burmak:


Alex-Burmak:
Replaces https://github.com/yandex/ch-tools/pull/150

PR #155, opened 2024-04-24
Replace print with logging

kirillgarbar:
- Almost all prints are replaced with logging call
- Stdout handler is added by default
- Method to disable stdout logger is added so it can be used with `quiet` option(like in `chadmin wait started`)

### DIFF
--- a/ch_tools/chadmin/internal/zookeeper.py
+++ b/ch_tools/chadmin/internal/zookeeper.py
@@ -112,7 +112,7 @@ def _set_node_value(zk, path, value):
         try:
             zk.set(path, value.encode())
         except NoNodeError:
-            logging.warning(f"Can not set for node: {path}  value : {value}")
+            logging.warning("Can not set for node: {}  value : {}", path, value)
 
 
 def _find_paths(zk, root_path, included_paths_regexp, excluded_paths_regexp=None):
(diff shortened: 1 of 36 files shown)


ch_tools/monrun_checks_keeper/main.py
```diff
@@ -66,6 +66,7 @@ def wrapper(ctx, *a, **kw):
                 status.set_code(1)
 
             log_message = f"Completed with {status.code};{status.message}"
+            logging.disable_stdout_logger()
```
kirillgarbar, line 69: Message with status code will be printed later in `status.report(ctx)`. Maybe we should replace `print` in `report` for logging and remove `log_status` from here.
Also, do we need to print exception info from the try block above? If not, we can disable stdout before logging exception.

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -493,7 +493,7 @@ def _get_zk_client(ctx):
         connect_str,
         auth_data=auth_data,
         timeout=timeout,
-        logger=logging.getLogger("chadmin"),
+        logger=logging.getNativeLogger("kazoo"),
```
kirillgarbar, line 496: Passing loguru logger to other libraries results in broken formatting due to `%s` usage. This way `InterceptHandler` passes correctly formatted message to the corresponding loguru logger.

ch_tools/monrun_checks_keeper/main.py
aalexfvk: Lets leave `print`. I think we can move `disable_stdout_logger` up.

aalexfvk (approved):

PR #156, opened 2024-04-25
Add optional traverse_shadow_remote_data_paths settings

aalexfvk:
Add optional traverse_shadow_remote_data_paths settings enabled for CH version greater or equal '24.2'
Add dynamic prefix when on_cluster is enabled.


ch_tools/chadmin/cli/object_storage_group.py
```diff
@@ -172,7 +174,17 @@ def _clean_object_storage(
...
-    prefix = object_name_prefix or disk_conf.prefix
+
+    if object_name_prefix is not None:
+        prefix = object_name_prefix
+    else:
+        prefix = disk_conf.prefix
+        # Yandex cloud specific
+        # Remove the last shard section from prefix
+        if on_cluster:
+            match = re.match(r"(.*)shard\d+/$", prefix)
```
Alex-Burmak, line 185: Shard name can be an arbitrary string and can mismatch `shard\d+`.

Suggest to replace the regexp with:
```
prefix = os.path.dirname(prefix)
```

And I guess it's not Yandex Cloud specific. We just imply that shard paths are on the same nesting level.

ch_tools/chadmin/cli/object_storage_group.py
```diff
@@ -172,7 +174,17 @@ def _clean_object_storage(
...
     """
     disk_conf: S3DiskConfiguration = ctx.obj["disk_configuration"]
-    prefix = object_name_prefix or disk_conf.prefix
+
+    if object_name_prefix is not None:
+        prefix = object_name_prefix
+    else:
+        prefix = disk_conf.prefix
+        # Yandex cloud specific
+        # Remove the last shard section from prefix
```
ianton-ru, line 183: Are you sure that make some specific behavior as default is a right way?
For example, in DC we name shards as `s#`, not `shard#`.

aalexfvk: Agree, this is a very questionable change. Removed it from this PR.

Alex-Burmak (approved):

PR #171, opened 2024-05-08
Cleanup support of old deprecated versions lower 22.8

Alex-Burmak:


MikhailBurdukov:
Should we remove it too? 
https://github.com/yandex/ch-tools/blob/25d90251482b8195a36e1f059cc64a5f26f38bf3/ch_tools/chadmin/internal/part.py#L159

MikhailBurdukov (approved):

PR #172, opened 2024-05-09
Remove useless multi_replicas flag.

k-morozov:
I suggest removing multi_replicas flag from replication_lag.

Right now we have a filter for total_replicas > 1 (because we don't calculate replication lag for 1 replica). So we don't use when total_replicas=1  (in all cases we check chart with if item.get("multi_replicas", False)). Instead 2 sql command we can use 1 with filter total_replicas>1.

### DIFF
--- a/ch_tools/common/commands/replication_lag.py
+++ b/ch_tools/common/commands/replication_lag.py
@@ -33,66 +33,61 @@ def estimate_replication_lag(ctx, xcrit, crit, warn, mwarn, mcrit, verbose=0):
             "Merges with 1000+ tries",
         ]
         for key, item in chart.items():
-            if item.get("multi_replicas", False):
-                tabletab = [
-                    key,
-                    item.get("delay", 0),
-                    item.get("tasks", 0),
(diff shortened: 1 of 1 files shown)


ch_tools/common/commands/replication_lag.py
```diff
@@ -33,66 +33,61 @@ def estimate_replication_lag(ctx, xcrit, crit, warn, mwarn, mcrit, verbose=0):
...
-                        if item.get("max_execution", 0)
-                        else 0
+            tabletab = [
+                key,
+                item.get("delay", 0),
+                item.get("tasks", 0),
+                item.get("max_execution", 0),
+                item.get("errors", 0),
+                item.get("user_fault", False),
+                item.get("re    tried_merges", 0),
```
Alex-Burmak, line 43: Incorrect field name. It should be `retried_merges` (without space after `re`).

Alex-Burmak:
LGTM except the comment above.

Alex-Burmak (approved):

PR #174, opened 2024-05-13
chadmin: --cluster option for "part-log list" command

Alex-Burmak:


MikhailBurdukov:
Is it possible to add a test?

Alex-Burmak:
> Is it possible to add a test?

Added a simply test that checks successful execution of chadmin commands including `part-log list`.

MikhailBurdukov (approved):

PR #181, opened 2024-05-15
Fix for set flag command

MikhailBurdukov:


ch_tools/chadmin/cli/table_group.py
```diff
@@ -715,6 +715,7 @@ def materialize_ttl_command(ctx, _all, dry_run, **kwargs):
     "--verbose",
     type=bool,
     is_flag=True,
+    default=True,
```
Alex-Burmak, line 718: It makes the behavior the same with and without `--verbose`. Probably we should remove this option as it's redundant now.

PR #182, opened 2024-05-16
Feature/drop detached merge tree table

k-morozov:


ch_tools/chadmin/cli/table_group.py
```diff
@@ -242,20 +243,41 @@ def columns_command(ctx, database_name, table_name):
...
 ):
     """
     Delete one or several tables.
     """
     cluster = get_cluster_name(ctx) if on_cluster else None
 
+    if detached:
+        logging.info("delete detached table")
+
+        database_name = kwargs.get("database_pattern")
```
Alex-Burmak, line 269: Suggest to add additional arguments for `database_name` and `table_name` to avoid using patterns allowing multi-selection as exact names. 

```
@argument("database_name", metavar="DATABASE", required=False)
@argument("table_name", metavar="TABLE", required=False)
```

Then deleting detached table will looks like 
```
table delete --detached test_db test_table
```

Also it would be good to forbid using inapplicable options when deleting detached tables. It can be done something like this: 
```
@constaint(If('detached', then=accept_none), ['database_pattern', 'exclude_database_pattern', ...
```
https://cloup.readthedocs.io/en/stable/pages/constraints.html

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +260,136 @@ def delete_table(
...
+        raise RuntimeError(
+            f"clickhouse-disks remove command has failed: retcode {code}, stderr: {stderr.decode()}"
+        )
+
+
+def delete_detached_table(ctx, database_name, table_name):
+    check_table_dettached(ctx, database_name, table_name)
+
+    local_metadata_table_path = (
+        CLICKHOUSE_METADATA_PATH + "/" + database_name + "/" + table_name + ".sql"
```
Alex-Burmak, line 363: ClickHouse escapes database and table names when constructing paths on file system. So this won't work for arbitrary names.

ch_tools/chadmin/internal/table.py
```diff
@@ -1,6 +1,17 @@
-from click import ClickException
+import os
+import uuid
 
+from click import ClickException, Context
+
+from ch_tools.chadmin.cli.data_store_group import (
```
MikhailBurdukov, line 6: Lets move these functions and constants into separate module. `ch_tools/chadmin/internal/clickhouse-disks.py` for example

ch_tools/chadmin/cli/data_store_group.py
```diff
@@ -15,6 +15,7 @@
 CLICKHOUSE_STORE_PATH = CLICKHOUSE_PATH + "/store"
 CLICKHOUSE_DATA_PATH = CLICKHOUSE_PATH + "/data"
 CLICKHOUSE_METADATA_PATH = CLICKHOUSE_PATH + "/metadata"
+S3_METADATA_STORE_PATH = CLICKHOUSE_PATH + "/disks/object_storage/store"
```
MikhailBurdukov, line 18: What if the disk name Is not `object_storage`? I think we should generate this variable in run-time like:
` CLICKHOUSE_PATH + f"/disks/{disk_name}/store"`

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +260,136 @@ def delete_table(
...
+            raise RuntimeError(f"No UUID in '{table_metadata_path}'")
+
+    assert _is_valid_uuid(table_uuid)
+
+    return table_uuid
+
+
+def check_table_dettached(ctx, database_name, table_name):
+    query = """
+        SELECT
```
MikhailBurdukov, line 297: I just wonder, is it possible to remove the detached table totally without queries to the clickhouse? For example for the scenario when clickhouse-server can't starts.  Will be useful such use-case?

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +260,136 @@ def delete_table(
...
+        return False
+
+    logging.info("There is an object storage.")
+    return True
+
+
+def _remove_table_data_from_object_storage(local_metadata_table_path):
+    table_uuid = _get_uuid_table(local_metadata_table_path)
+
+    object_storage_table_data_path = "store" + "/" + table_uuid[:3] + "/" + table_uuid
```
MikhailBurdukov, line 339: It's on you, but I like more:
```suggestion
    short_uuid = table_uuid[:3]
    object_storage_table_data_path = "store" + "/" + short_uuid + "/" + table_uuid
```

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +260,136 @@ def delete_table(
...
+
+    logging.info(
+        "Table has UUID {}, path in S3: {}.", table_uuid, object_storage_table_data_path
+    )
+
+    disk = "object_storage"
+    disk_config_path = make_ch_disks_config(disk)
+
+    code, stderr = remove_from_ch_disk(
```
MikhailBurdukov, line 348: The table can be present on several disks simultaneously. 
For example with the storage_policy:
```
<default>
      <move_factor>0.01</move_factor>
      <volumes>
          <default>
              <disk>default</disk>
          </default>
          <object_storage>
              <disk>object_storage</disk>
          </object_storage>
      </volumes>
</default>

```

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +260,136 @@ def delete_table(
...
+        ctx,
+        query,
+        database_name=database_name,
+        table_name=table_name,
+        format_="JSON",
+    )["data"]
+
+    if 0 != len(response):
+        raise RuntimeError(
+            f"Table '{database_name}'.'{table_name}' is attach. Use delete without --detach flag."
```
Alex-Burmak, line 314: `is attach` -> `is attached`.

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +260,136 @@ def delete_table(
...
+        )
+
+
+def _has_object_storage(ctx: Context) -> bool:
+    query = """
+        SELECT 1 FROM system.disks WHERE name = 'object_storage'
```
Alex-Burmak, line 320: It's better to not depend on concrete disk names. Suggest to select non-local disks by disk type and perform cleanup for them using `clickhouse-disks` tool. Probably we can do the same for local disks too.

ch_tools/chadmin/cli/table_group.py
```diff
@@ -242,20 +243,41 @@ def columns_command(ctx, database_name, table_name):
     default=False,
     help="Enable dry run mode and do not perform any modifying actions.",
 )
+@option(
+    "--detached",
+    is_flag=True,
+    help="Delete detached tables (with nonreplicated engine).",
```
Alex-Burmak, line 249: Suggest to add a check that deleting table has supported table engine.

Alex-Burmak (commented):
Suggest to add a test for new feature.

ch_tools/chadmin/cli/table_group.py
k-morozov: Thank you, good things.

ch_tools/chadmin/internal/table.py
k-morozov: Ok. (filename clickhouse_disks.py)

ch_tools/chadmin/cli/data_store_group.py
k-morozov: It is just a refactoring. We use object_storage [here](https://github.com/yandex/ch-tools/pull/182/files/127dcbfcd73cf8d5890c1be50ac1d7d2744404dc#diff-12dd2989ce7a369b8d062c461ed0d7b0cdb43bb7057f031526cebc6c6f689bebL191)

Let's discuss this in a different PR if you don't mind?

ch_tools/chadmin/internal/table.py
k-morozov: Ok, Im going to remove logic that relates with disk name.

ch_tools/chadmin/internal/table.py
k-morozov: And what?

ch_tools/chadmin/internal/table.py
k-morozov: > is it possible to remove the detached table totally without queries to the clickhouse
No.
Firstly we need make sure that table is realy detached so we need send a querry. 

> Will be useful such use-case?
I think no. If clickhouse-server doesn't start, its mean that problem doesn't relating with detached table and need to fix another problem.

k-morozov:
> Suggest to add a test for new feature.

ok

ch_tools/chadmin/internal/clickhouse_disks.py
```diff
@@ -0,0 +1,52 @@
...
+    cmd = f"clickhouse-disks { '-C ' + disk_config_path if disk_config_path else ''} --disk {disk} remove {path}"
+    logging.info("Run : {}", cmd)
+
+    proc = subprocess.run(
+        cmd, shell=True, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    )
+    return (proc.returncode, proc.stderr)
+
+
+def remove_from_disk(path):
```
MikhailBurdukov, line 45: Suggest to keep in this file only functionality related with `clickhouse-disks` cmd tool.

We have used this function as workaround, due to poor performance of `clickhouse-disks remove` in case when there are too many files to remove(>1000000).

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +261,173 @@ def delete_table(
...
+
+def check_table_engine_supported(table_metadata_path: str) -> None:
+    engine = _get_engine_table(table_metadata_path)
+    logging.info("Found engine {} in metadata {}", engine, table_metadata_path)
+
+    if "Replic" in engine:
+        raise RuntimeError(
+            f"Doesn't support removing detached table with engine {engine}"
+        )
+
```
MikhailBurdukov, line 338:
```suggestion
if "MergeTree" not in engine:
   raise ......
```

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +261,173 @@ def delete_table(
...
+    if not os.path.exists(local_metadata_table_path):
+        raise RuntimeError(
+            f"No metadata file for table '{escaped_database_name}'.'{escaped_table_name}' by path {local_metadata_table_path}."
+        )
+
+    check_table_engine_supported(local_metadata_table_path)
+
+    table_uuid = _get_uuid_table(local_metadata_table_path)
+    for disk in _get_disk_names(ctx):
```
MikhailBurdukov, line 409: Fyi There are some bugs in `23.3-23.8` versions with loading  cache disks in `clickhouse-disks`, so I think it make sense to filter it.

ch_tools/chadmin/internal/table.py
MikhailBurdukov: Sorry, miss-understanding..

Is `_remove_table_data_from_disk` should remove from any disk not only from object_storage? It seems like we will remove from default disk too? Right? 
https://github.com/yandex/ch-tools/blob/9b97d2fb767f051f0e72982ba28a2c975486bed1/ch_tools/chadmin/internal/table.py#L409-L410

So naming in this function just had confused me...
https://github.com/yandex/ch-tools/blob/9b97d2fb767f051f0e72982ba28a2c975486bed1/ch_tools/chadmin/internal/table.py#L363

rename?

k-morozov: After  [suggestion](https://github.com/yandex/ch-tools/pull/182#discussion_r1603632447) I propose to use clickhouse-disks for removing from default too. Thank you, I skipped deprecation name object_storage_table_data_path. Im going to rename to just table_data_path and update log message.

ch_tools/chadmin/internal/table.py
k-morozov: Yea, I had saw a problem with cache so I added a filter [here](https://github.com/yandex/ch-tools/pull/182/files/9b97d2fb767f051f0e72982ba28a2c975486bed1#diff-e5b2e346ca78f718d722d340262e7567c9b38626bc7b902915434df10b8f098eR344).

ch_tools/chadmin/internal/clickhouse_disks.py
k-morozov: Ok, Im removing it to utils.

ch_tools/chadmin/internal/table.py
MikhailBurdukov: Okey.
Let's do some refactoring, because it's not obvious for me that `_get_disks_names` will return all disks except 'object_storage_cache'.
 
Propose something like:
```
def _get_disk_names(ctx):
       query = "SELECT name,type FROM system.disks"

....
for disk in filter(_get_disk_names(ctx), /*remove cache disks from the list*/):
```
or 
```
def _get_disks_names(ctx, types_to_exclude):
       type_to_exclude_ch_array = /*convert types_to_exclude to string in clickhouse array format */
       query = "SELECT name,type FROM system.disks WHERE not has({type_to_exclude_ch_array}, type)"

```

ch_tools/chadmin/cli/table_group.py
```diff
@@ -242,20 +252,49 @@ def columns_command(ctx, database_name, table_name):
...
+    table_name,
     **kwargs,
 ):
     """
     Delete one or several tables.
     """
     cluster = get_cluster_name(ctx) if on_cluster else None
 
+    if detached:
+        delete_detached_table(
```
Alex-Burmak, line 291: Suggest to implement support of dry-run mode for deleting detached tables. Or at least mark --dry-run option as incompatible with --detached.

ch_tools/chadmin/cli/table_group.py
```diff
@@ -242,20 +252,49 @@ def columns_command(ctx, database_name, table_name):
...
+    If(IsSet("detached"), then=require_all),
+    [
+        "database_name",
+        "table_name",
+    ],
+)
+@constraint(
+    If(IsSet("detached"), then=accept_none),
+    [
+        "on_cluster",
```
Alex-Burmak, line 270: The list looks incomplete. For example, --sync is missing.

ch_tools/chadmin/cli/table_group.py
```diff
@@ -220,14 +230,14 @@ def columns_command(ctx, database_name, table_name):
...
-    constraint=RequireAtLeast(1),
+    constraint=If(IsSet("detached"), then=accept_none, else_=RequireAtLeast(1)),
 )
 @option(
     "--cluster",
     "--on-cluster",
     "on_cluster",
     is_flag=True,
-    help="Delete tables on all hosts of the cluster.",
+    help="Delete tables on all hosts of the cluster. Doesn't work with --detached flag.",
```
Alex-Burmak, line 240: > "Doesn't work with --detached flag." 

Is it required? Don't automatic help messages cover option constraints and dependencies?

ch_tools/chadmin/cli/table_group.py
```diff
@@ -167,6 +175,8 @@ def columns_command(ctx, database_name, table_name):
 
 
 @table_group.command("delete")
+@argument("database_name", metavar="DATABASE", required=False)
```
Alex-Burmak, line 178: Suggest to implement support of database name and table name args for non-detached tables.

tests/features/drop_detached_table.feature
```diff
@@ -0,0 +1,80 @@
...
+    """
+    """
+    When we execute command on clickhouse01
+    """
+    ls -l /var/lib/clickhouse/data/test_drop_detach_db/
+    """
+    Then we get response
+    """
+    total 0
+    """
```
Alex-Burmak, line 80: Suggest to also check that data from s3 is deleted too.

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +263,193 @@ def delete_table(
...
+        table_data_path,
+    )
+
+    disk_config_path = make_ch_disks_config(disk)
+
+    if equal_ch_version(
+        ctx, version="22.8.21.38"
+    ) and not _is_should_use_ch_disk_remover(disk, table_data_path):
+        logging.info(
+            "Dir not exists. Skip launch clickhouse-disks for Clickhouse 22.8."
```
MikhailBurdukov, line 389: Lets add a path which we keep on the disk, because when we remove metadata files, it will be hard to get `uuid` and match with path in `store`

tests/features/drop_detached_table.feature
```diff
@@ -0,0 +1,80 @@
...
+  Scenario: Drop detached table from object_storage
+    Given we have executed queries on clickhouse01
+    """
+    CREATE TABLE IF NOT EXISTS test_drop_detach_db.test_table_object_storage (n Int32)
+    ENGINE = MergeTree
+    ORDER BY n
+    SETTINGS storage_policy = 'object_storage';
+
+    INSERT INTO test_drop_detach_db.test_table_object_storage (n) SELECT number FROM system.numbers LIMIT 10;
+    DETACH TABLE test_drop_detach_db.test_table_object_storage SYNC;
```
MikhailBurdukov, line 56: To cover all usage scenarios, lets add a test with `detach permanently`

ch_tools/chadmin/cli/table_group.py
k-morozov: I will mark dry as incompatible with --detached.

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +263,193 @@ def delete_table(
...
+
+
+def _get_uuid_table(table_metadata_path: str) -> str:
+    table_uuid = _get_data_table(table_metadata_path, pattern="UUID", offset=1)
+    assert _is_valid_uuid(table_uuid)
+
+    return table_uuid
+
+
+def _get_engine_table(table_metadata_path: str) -> str:
```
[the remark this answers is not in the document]
k-morozov: Ye, it makes sense.

ch_tools/chadmin/cli/table_group.py
k-morozov: I suggest to do it in another PR, if you don't mind.

ch_tools/chadmin/cli/table_group.py
k-morozov: You right. I will remove it.

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +264,205 @@ def delete_table(
...
+def _get_table_data(table_metadata_path: str, pattern: str, offset: int) -> str:
+    result = ""
+
+    with open(table_metadata_path, "r", encoding="utf-8") as metadata_file:
+        for line in metadata_file:
+            if pattern in line:
+                parts = line.split()
+                try:
+                    index = parts.index(pattern)
+                    # check size?
```
MikhailBurdukov, line 297: ?

ch_tools/chadmin/internal/table.py
```diff
@@ -249,6 +264,205 @@ def delete_table(
...
+
+    response = execute_query(
+        ctx,
+        query,
+        database_name=database_name,
+        table_name=table_name,
+        format_="JSON",
+    )["data"]
+
+    if 0 != len(response):
```
MikhailBurdukov, line 327:
```suggestion
    if len(response):
```

MikhailBurdukov (approved):

ch_tools/chadmin/cli/table_group.py
```diff
@@ -242,20 +252,51 @@ def columns_command(ctx, database_name, table_name):
...
+    "--detached",
+    is_flag=True,
+    help="Delete detached tables (with nonreplicated engine).",
+)
+@constraint(
+    If(IsSet("detached"), then=require_all),
+    [
+        "database_name",
+        "table_name",
+        "sync_mode",
```
Alex-Burmak, line 265: Shouldn't it be in `accept_none` group?

Alex-Burmak (approved):

ch_tools/chadmin/cli/table_group.py
k-morozov: No, sync_mode has default=True and remove works sync. There is no support for async drop detached table.

ch_tools/chadmin/internal/table.py
k-morozov: Up [here](https://github.com/yandex/ch-tools/pull/184)

ch_tools/chadmin/internal/table.py
k-morozov: Up [here](https://github.com/yandex/ch-tools/pull/184)

PR #183, opened 2024-05-21
Remove ddl tasks in cleanup-removed-hosts-metadata command

MikhailBurdukov:


ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -386,38 +387,97 @@ def get_host_mention_in_task(zk, host, ddl_task_path):
...
+        for _ in range(0, max_attemps):
+            all_tasks_finished = True
+            for ddl_data in dll_tasks_to_remove:
+                ddl_path = ddl_data[0]
+                ddl_size = ddl_data[1]
+                if not _check_ddl_task_finished(zk, ddl_size, ddl_path):
+                    all_tasks_finished = False
+                    break
+
+            if all_tasks_finished:
```
aalexfvk, line 426: We can [use](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops) `else:` after `for` loops instead of a flag.

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -386,38 +387,97 @@ def get_host_mention_in_task(zk, host, ddl_task_path):
...
+
+        return ddl_number_of_hosts == len(
+            _get_children(zk, os.path.join(ddl_path, "finished/"))
+        )
+
+    def _ensure_ddl_tasks_finished(zk, dll_tasks_to_remove):
+        """
+        Go thought list of tasks to remove and tries to wait until all of them finish.
+        """
+        wait_time = 0.3
```
aalexfvk, line 414: Let's make this vars as function parameters with default values

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -386,38 +387,97 @@ def get_host_mention_in_task(zk, host, ddl_task_path):
...
+        Check that ddl is finished.
+        """
+        if not zk.exists(ddl_path):
+            return True
+
+        return ddl_number_of_hosts == len(
+            _get_children(zk, os.path.join(ddl_path, "finished/"))
+        )
+
+    def _ensure_ddl_tasks_finished(zk, dll_tasks_to_remove):
```
aalexfvk, line 410: `_wait_ddl_tasks_finished` ?

ch_tools/chadmin/internal/zookeeper.py
aalexfvk: I meant that we can get rid of `all_tasks_finished` at all, doing early return

aalexfvk (approved):

PR #184, opened 2024-05-22
delete particular table.

k-morozov:
Previously conversation https://github.com/yandex/ch-tools/pull/182#discussion_r1607804879


ch_tools/chadmin/cli/table_group.py
```diff
@@ -297,7 +323,26 @@ def delete_command(
...
+            table_name=table_name,
+            cluster=cluster,
+            sync_mode=sync_mode,
+            echo=True,
+            dry_run=dry_run,
+        )
+        return
+
+    tables = list_tables(ctx, **kwargs)
+    if 0 == len(tables):
```
MikhailBurdukov, line 340:
```suggestion
    if len(tables):
```

ch_tools/chadmin/cli/table_group.py
```diff
@@ -297,7 +323,26 @@ def delete_command(
         )
         return
 
-    for t in list_tables(ctx, **kwargs):
+    if database_name is not None and table_name is not None:
```
MikhailBurdukov, line 326:
```suggestion
    if database_name and table_name:
```

tests/features/table_delete.feature
```diff
@@ -0,0 +1,261 @@
+Feature: chadmin delete detached table commands
```
MikhailBurdukov, line 1: Also lets add a test to cover cli options usage, suggest to use `Scenario Outline` for that.

ch_tools/chadmin/cli/table_group.py
k-morozov: Why? It will be another result.

MikhailBurdukov: Sorry 
```suggestion
    if not len(tables):
```

tests/features/table_delete.feature
k-morozov: I think your suggestion is mostly relate with test coverage of existing functionality (table delete by pattern) than current adding feature of deleting particular table. So we have a lot of variants of using args/options, checking constraints when we use command table delete. I suppose it's better to do it in a separate task, if you dont mind.

MikhailBurdukov (approved):

PR #187, opened 2024-05-23
Support drop replicated detached table

k-morozov:
I suggest to move all logic relating with parse from table metadata in on class.
up: Also support drop replicated detached table.


ch_tools/chadmin/cli/table_metadata.py
```diff
@@ -0,0 +1,93 @@
...
+        return str(val) == uuid_str
+
+    @staticmethod
+    def _is_valid_engine(engine: str) -> bool:
+        return engine in PARSER_SUPPORT_TABLE_ENGINE
+
+    @staticmethod
+    def _parse_uuid(line: str) -> str:
+        result = ""
+        parts = line.split()
```
MikhailBurdukov, line 67: I think the implementation will be much easier if we will use regex here.

ch_tools/chadmin/cli/table_metadata.py
```diff
@@ -0,0 +1,93 @@
...
+        try:
+            index = parts.index(UUID_PATTERN)
+            if len(parts) < index + OFFSET_UUID_PATTERN + 1:
+                raise RuntimeError("Failed parse UUID from metadata.")
+
+            result = parts[index + OFFSET_UUID_PATTERN].strip("'")
+        except ValueError:
+            raise RuntimeError(f"Failed parse {UUID_PATTERN} from metadata.")
```
MikhailBurdukov, line 75: do not throw if we can't parse uuid, just return None + warning msg.
User might have it in as column name:
```
CREATE TABLE test
(
    `_UUID` UInt8
)
ENGINE = MergeTree
ORDER BY _UUID

```

ch_tools/chadmin/cli/table_metadata.py
```diff
@@ -0,0 +1,93 @@
...
+    "ReplacingMergeTree",
+    "SummingMergeTree",
+    "AggregatingMergeTree",
+    "CollapsingMergeTree",
+    "VersionedCollapsingMergeTree",
+    "ReplicatedMergeTree",
+]
+
+
+class TableMetadata:
```
MikhailBurdukov, line 19: This class breaks `single responsibility`. Suggest refactoring:
```
class TableMetadata -> class which only keeps parsed metadata.
class MetadataParser -> class responsible for metadata parsing.
```

ch_tools/chadmin/cli/table_metadata.py
```diff
@@ -0,0 +1,93 @@
...
+        if self._table_uuid is None:
+            raise RuntimeError(f"Empty UUID from metadata: '{table_metadata_path}'")
+
+        if self._table_engine is None:
+            raise RuntimeError(
+                f"Empty table engine from metadata: '{table_metadata_path}'"
+            )
+
+    @staticmethod
+    def _is_valid_uuid(uuid_str: str) -> bool:
```
MikhailBurdukov, line 53: I can't imagine the case when we will need it. Lets make it non-static by now

ch_tools/chadmin/cli/table_metadata.py
```diff
@@ -0,0 +1,93 @@
+import re
+import uuid
+
+UUID_PATTERN = "UUID"
+OFFSET_UUID_PATTERN = 1
+ENGINE_PATTERN = "ENGINE"
+
+PARSER_SUPPORT_TABLE_ENGINE = [
```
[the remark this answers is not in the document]
k-morozov: I changed it to enum

ch_tools/chadmin/cli/table_metadata.py
k-morozov: It is a way to check that metadata doesn't broken. Its important because all next removing basing on UUID from metadata file. We use it only inside the class and so it makes sense yo use as static.

ch_tools/chadmin/cli/table_metadata.py
k-morozov: Good thought. I think TableMetadataParser is better.

ch_tools/chadmin/cli/table_metadata.py
k-morozov: We must finish work if don't find UUID because all removing logic in the next steps based on uuid. I will update parse uuid logic for fix your scenarious.

MikhailBurdukov: >We must finish work if don't find UUID because all removing logic in the next steps based on uuid
Lest check it in removing logic

ch_tools/chadmin/cli/table_metadata.py
```diff
@@ -0,0 +1,93 @@
...
+
+        with open(table_metadata_path, "r", encoding="utf-8") as metadata_file:
+            for line in metadata_file:
+                if UUID_PATTERN in line:
+                    self._table_uuid = TableMetadata._parse_uuid(line)
+                    continue
+                if ENGINE_PATTERN in line:
+                    result = TableMetadata._parse_engine(line)
+                    self._table_engine = result
+                    break
```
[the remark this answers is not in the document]
k-morozov: Why? We stable format.

ch_tools/chadmin/cli/table_metadata.py
```diff
@@ -0,0 +1,93 @@
...
+
+        if self._table_uuid is None:
+            raise RuntimeError(f"Empty UUID from metadata: '{table_metadata_path}'")
```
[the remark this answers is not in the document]
k-morozov: I suggest to throw an exception for simular cases because delete based on uuid. So we cannt remove this detached table without uuid.

ch_tools/chadmin/cli/table_metadata.py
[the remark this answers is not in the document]
MikhailBurdukov: Because it makes logic more complicated. We have to alway keep in mind what is the order. 
Also what if we will want several keys from single line and both of them are optional?
```
... (optional) key1 value1 .... (optional) key2 value2
``` 
Such case will be complicated to handle correctly with `continue - break` logic

ch_tools/chadmin/cli/table_metadata.py
k-morozov: > Why if the method is used only inside the class, then it should be static?

We don't use self so we dont have any reasons to make method nonstatic.

ch_tools/chadmin/cli/table_metadata.py
k-morozov: Currently, if we create a class TableMetadataParser, there is a guarantee that the data is consistent and correct. If we move the UUID check as you suggested, it would break this guarantee. I would like to avoid such a scenario.

ch_tools/chadmin/internal/table.py
```diff
@@ -450,14 +401,19 @@ def delete_detached_table(ctx, database_name, table_name):
...
-    table_uuid = _get_uuid_table(local_metadata_table_path)
     for disk_type, disk_name in _get_disks_data(ctx).items():
         _remove_table_data_from_disk(
-            table_uuid=table_uuid, disk_name=disk_name, disk_type=disk_type
+            table_uuid=table_metadata.table_uuid,
+            disk_name=disk_name,
+            disk_type=disk_type,
         )
 
+    if table_metadata.table_engine == MergeTreeFamilyEngines.REPLICATED_MERGE_TREE:
```
MikhailBurdukov, line 413: table_metadata.is_replicated()

ch_tools/chadmin/cli/table_metadata.py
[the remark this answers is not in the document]
k-morozov: Its doesnt matter right now. I will remove order so we will descuss about it later.

ch_tools/chadmin/cli/table_metadata_parser.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @property
+    def table_engine(self):
+        return self._table_engine
+
+    def __init__(self, table_metadata_path: str):
+        assert table_metadata_path.endswith(".sql")
+
+        self._table_uuid = None
+        self._table_engine = None
```
MikhailBurdukov, line 41: still broken single resp
```suggestion
class TableMetadata:
       self.table_uuid = None
       self.table_engine = None
       

class TableMetadataParser:
    @property
    def get_table_metadata(self):
        return self._table_metadata

    def __init__(self, table_metadata_path: str):
        assert table_metadata_path.endswith(".sql")
        self._table_metadata
```

k-morozov: If we are talking about the Single Responsibility Principle, the above option does not significantly differ from the original. At the very least, it still stores the same data, but with additional abstraction. In my opinion, a parser is not the type of task where this principle needs to be discussed. I know plenty of examples of parsers that either store or do not store data themselves. From the perspective of solving the current task, this is not important.

In any case, I have now implemented the classic approach.

ch_tools/chadmin/cli/table_metadata_parser.py
```diff
@@ -0,0 +1,99 @@
...
+            raise RuntimeError("Failed parse UUID from metadata.")
+        return result
+
+    @staticmethod
+    def _parse_engine(line: str) -> MergeTreeFamilyEngines:
+        pattern = re.compile(r"ENGINE = (\w+)")
+
+        match = pattern.search(line)
+        if not match:
+            raise RuntimeError(f"Failed parse {ENGINE_PATTERN} from metadata.")
```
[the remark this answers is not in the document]
k-morozov: We are discussing whether it is better to return None or raise exceptions. This is a contentious topic, and I do not see any clear arguments for why we should change aproach.

tests/features/table_delete.feature
```diff
@@ -258,4 +258,119 @@ Feature: chadmin delete detached table commands
...
+    """
+    """
+    Then the list of children on clickhouse01 for zk node /clickhouse/tables/{shard}/ are equal to
+    """
+    """
+
+  Scenario: Drop replicated table from some hosts
+    Given we have executed queries on clickhouse02
+    """
+    CREATE DATABASE IF NOT EXISTS test_drop_db;
```
MikhailBurdukov, line 299: ON CLUSTER ?

ch_tools/chadmin/internal/table.py
```diff
@@ -450,14 +399,25 @@ def delete_detached_table(ctx, database_name, table_name):
...
+    if table_metadata.table_engine.is_table_engine_replicated():
+        current_node = table_metadata.replica_path + "/replicas/"
+        current_replica = current_node + table_metadata.replica_name
+
+        logging.info("Remove replica: {}", current_replica)
+        delete_zk_node(ctx, current_replica)
+        if not list_zk_nodes(ctx, current_node):
+            logging.info("No replicas, remove node: {}", current_node)
+            delete_zk_node(ctx, table_metadata.replica_path)
+
```
MikhailBurdukov, line 420: I think better to use `clean_zk_metadata_for_hosts`, because logic of removing host might be more comlicated sometimes: we want to remove all replica name mentioning from the zookeeper, not only from `/replicas/`, and if the replica is last in table, we have to remove whole table path.

ch_tools/chadmin/cli/table_metadata_parser.py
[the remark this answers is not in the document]
MikhailBurdukov: probably u right. Let keep it as is.

ch_tools/chadmin/internal/table.py
```diff
@@ -450,14 +399,25 @@ def delete_detached_table(ctx, database_name, table_name):
...
     for disk_type, disk_name in _get_disks_data(ctx).items():
         _remove_table_data_from_disk(
-            table_uuid=table_uuid, disk_name=disk_name, disk_type=disk_type
+            table_uuid=table_metadata.table_uuid,
+            disk_name=disk_name,
+            disk_type=disk_type,
         )
 
+    if table_metadata.table_engine.is_table_engine_replicated():
+        current_node = table_metadata.replica_path + "/replicas/"
```
MikhailBurdukov, line 412: I think better to use clean_zk_metadata_for_hosts, because it is duplicating same logic, also logic of removing host might be more complicated sometimes: we want to remove all replica name mentioning from the zookeeper, not only from /replicas/.

tests/features/table_delete.feature
MikhailBurdukov, line 299: ON CLUSTER

ch_tools/chadmin/cli/table_metadata_parser.py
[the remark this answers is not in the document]
MikhailBurdukov: Probably you right. Lets keep it as is

ch_tools/chadmin/internal/table.py
```diff
@@ -450,12 +415,27 @@ def delete_detached_table(ctx, database_name, table_name):
...
     for disk_type, disk_name in _get_disks_data(ctx).items():
         _remove_table_data_from_disk(
-            table_uuid=table_uuid, disk_name=disk_name, disk_type=disk_type
+            table_uuid=table_metadata.table_uuid,
+            disk_name=disk_name,
+            disk_type=disk_type,
+        )
+
+    if table_metadata.table_engine.is_table_engine_replicated():
+        fqdn = _get_fqdn(ctx)
```
MikhailBurdukov, line 428: Can we get host name from macros list, to not perform extra query.

k-morozov: Yea, lets use replica_name.

MikhailBurdukov (approved):

PR #189, opened 2024-05-25
Support of database_name and table_name args for replica management commands

Alex-Burmak:


ch_tools/chadmin/cli/replica_group.py
```diff
@@ -222,13 +253,26 @@ def restart_replica_command(ctx, on_cluster, dry_run, **kwargs):
...
+):
     """
     Restore one or several table replicas.
     """
     cluster = get_cluster_name(ctx) if on_cluster else None
-    ro_replicas = list_table_replicas(ctx, is_readonly=True, **kwargs)
-    for replica in ro_replicas:
+
+    if database_name and table_name:
+        replicas = [get_table_replica(ctx, database_name, table_name)]
```
k-morozov, line 271: When we use get_table_replica we don't specify RO flag but in list_table_replicas we use this flag.
Is it ok to restore non-readonly replica later?

Alex-Burmak: There is a check in ClickHouse:
```
SYSTEM RESTORE REPLICA _system.write_sli_part

Received exception from server (version 24.3.3):
Code: 36. DB::Exception: Received from localhost:9000. DB::Exception: Replica must be readonly. (BAD_ARGUMENTS)
```

In addition, I added the check to the tool itself.
```
# chadmin replica restore _system write_sli_part
Error: Replicated table `_system`.`write_sli_part` must be in read-only in order to perform restore.
```

k-morozov (approved):

PR #193, opened 2024-06-10
S3 orphaned objects monitoring rewrite

NikitaUnisikhin:


ch_tools/chadmin/cli/object_storage_group.py
```diff
@@ -119,6 +128,11 @@ def clean_command(
         keep_paths,
         use_saved_list,
     )
+
+    if total_size_to_file:
+        with open(ORPHANED_OBJECTS_LOG, mode="w+", encoding="utf-8") as file:
+            file.write(str(total_size))
```
Alex-Burmak, line 134: For better extendability, suggest to use json or alternative data format.

It can be something like this:
```
{
  "orphaned_objects_size": 1024
}
```

Related suggestions:
- Use `/tmp/object_storage_cleanup_state.json` instead of `/var/log/s3_orphaned_objects.log`
- Rename option `--total-size-to-file` to `--store-state`

NikitaUnisikhin: thanks for the comment, made edits

Alex-Burmak (approved):

PR #194, opened 2024-06-11
update tests for drop detached tables

k-morozov:


tests/steps/clickhouse.py
```diff
@@ -86,3 +92,71 @@ def step_put_config(context, path, node):
...
+    query = f"SELECT uuid FROM system.tables WHERE name='{table}'"
+    ret_code, response = get_response(context, node, query)
+    assert 200 == ret_code
+
+    if not hasattr(context, "uuid_to_table"):
+        context.uuid_to_table = {}
+    context.uuid_to_table[table] = response
+
+
+def _get_uuid_dirs_with_same_prefix_name(context, table_uuid, disk, node):
```
MikhailBurdukov, line 108: lets move `_get_uuid_dirs_with_same_prefix_name` and `_get_table_uuid ` to `integration/modules/clickhouse.py`. Better to keep here only step functions.

MikhailBurdukov, line 108: And this method looks too specific. Can we just check the dir with such uuid instead of getting the full list with common prefix?  Like `check_table_uuid_dir(context, uuid, disk, node)`

k-morozov: Ok, I dont mind.

MikhailBurdukov (approved):

PR #196, opened 2024-06-18
MDB-28831 Use parallel aproach for restore replica

MikhailBurdukov:


ch_tools/chadmin/cli/replica_group.py
```diff
@@ -278,47 +290,67 @@ def restore_command(
     else:
         replicas = list_table_replicas(ctx, is_readonly=True, **kwargs)
 
+    rotines: List[WorkerTask] = []
     for replica in replicas:
```
Alex-Burmak, line 294: `rotines` here and `tasks` in `data_store_group.py`. Why is it used different names for the same thing?

MikhailBurdukov: Fixed

Alex-Burmak (approved):

PR #203, opened 2024-07-02
MDB-29707: Better filtering and more logs for clean_zk_metadata_for_hosts

MikhailBurdukov:


ch_tools/chadmin/cli/zookeeper_group.py
```diff
@@ -303,14 +303,24 @@ def delete_ddl_task_command(ctx, tasks):
     name="cleanup-removed-hosts-metadata",
     help="Remove metadata from Zookeeper for specified hosts.",
 )
+@option(
+    "--clean-ddl-queue/--not-clean-ddl-queue",
```
Alex-Burmak, line 307: Suggest to replace `not` with `no` for consistency with other options.

A couple of examples from ch-tools repo:
```
    "--verify-ssl-certs/--no-verify-ssl-certs",
    "--ensure-monitoring-user/--no-ensure-monitoring-user",
```

Alex-Burmak (approved):

PR #204, opened 2024-07-03
CHINA-34: add config controls for geobase and orphaned-objects checks

Nuzhnov:


Alex-Burmak:
Suggest to use the solution from https://github.com/yandex/ch-tools/pull/206 for disabling the geobase check.

In the rest - LGTM

ch_tools/common/clickhouse/config/storage_configuration.py
```diff
@@ -3,7 +3,6 @@
 
 # YC specific value for sanity checking
 # TODO: pass bucket, prefix, endpoint as arguments to cli for versatility
```
ekopanev, line 5: seems like we can delete this comment too?

Alex-Burmak (approved):

PR #210, opened 2024-07-10
Log command start and finish

kirillgarbar:
To configure logging for each command and group in chadmin custom command group had to be added, like for monrun_checks and monrun_checks_keeper.

### DIFF
--- a/ch_tools/chadmin/cli/chadmin_group.py
+++ b/ch_tools/chadmin/cli/chadmin_group.py
@@ -60,3 +60,17 @@ class Chadmin(cloup.Group):
             section=section,
             fallback_to_default_section=fallback_to_default_section,
         )
+
+    def add_group(
+        self,
+        cmd: click.Group,
+        name: Optional[str] = None,
+        section: Optional[cloup.Section] = None,
+        fallback_to_default_section: bool = True,
+    ) -> None:
+        super().add_command(
+            cmd,
+            name=name,
+            section=section,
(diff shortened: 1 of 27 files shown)


ch_tools/chadmin/cli/chadmin_group.py
```diff
@@ -0,0 +1,62 @@
...
+                cmd.name,
+                {
+                    **ctx.parent.params,
+                    **ctx.params,
+                },
+                ctx.args,
+                __version__,
+            )
+
+            result = cmd_callback(*a, **kw)
```
MedvedewEM, line 50: Do we log something in case of error/fail in `cmd_callback`?
Maybe add some common message like in `monrun_checks_keeper` [here](https://github.com/yandex/ch-tools/pull/210/files#diff-073ad2263dfe496edc132c2c57a8d2d146262e3e5da43bbd02d8c0a5b1c4e51aR73)?

kirillgarbar:
Also fixed messages like `[DEBUG   ] chadmin: started: Executing command 'wait'` instead of `[DEBUG   ] chadmin: started: Executing command 'started'` because we added command groups to another group but with the same `add_command` method

ch_tools/chadmin/cli/chadmin_group.py
aalexfvk: It was wrapped in `try...except`

aalexfvk (approved):

PR #219, opened 2024-08-08
Store state data in zookeeper

NikitaUnisikhin:


ch_tools/chadmin/cli/object_storage_group.py
```diff
@@ -101,10 +105,10 @@ def object_storage_group(ctx: Context, disk_name: str) -> None:
...
 )
 @option(
-    "--store-state",
-    "store_state",
-    is_flag=True,
-    help=("Write total size of orphaned objects to log file."),
+    "--store-state-path",
+    "store_state_path",
+    default="",
+    help=("Zookeeper node path for storage total size of orphaned objects."),
```
Alex-Burmak, line 111: I think we need both options - store state in local file and store state in ZooKeeper. There are ClickHouse clusters with and without ZooKeeper.

ch_tools/monrun_checks/ch_orphaned_objects.py
```diff
@@ -4,12 +4,24 @@
...
+    "--store-state-local",
+    "store_state_local",
+    is_flag=True,
+    help="Get total size of orphaned objects from local file.",
+)
+@click.option(
+    "--store-state-zk-path",
+    "store_state_zk_path",
+    help="Zookeeper node path from which the total size of orphaned objects will be taken.",
+)
```
Alex-Burmak, line 24: It looks a bit strange that these options start with `--store` as the command only reads the state. Suggest to rename the options to `--state-local` and `--state-zk-path` (or to something else without `--store` suffix).

Alex-Burmak:
Is the PR in draft on purpose? Or you just forgot to click "Ready for review"?

ch_tools/monrun_checks/ch_orphaned_objects.py
NikitaUnisikhin: thank, done

ch_tools/chadmin/cli/object_storage_group.py
NikitaUnisikhin: thanks, done

NikitaUnisikhin:
Yes, I forgot to click "Ready for review"

By the way, what do you think about making the store-state-zk flag and creating a default path for storing statistics? I don’t like  that in this case there will be one path for storing statistics for hosts from different shards, but perhaps this problem should be left on the user’s side. But in some scenarios it may be useful.

Alex-Burmak:
> By the way, what do you think about making the store-state-zk flag and creating a default path for storing statistics? I don’t like that in this case there will be one path for storing statistics for hosts from different shards, but perhaps this problem should be left on the user’s side. But in some scenarios it may be useful.

I have no strong preference between the current implementation and described rework. Choose the one you prefer.

> It would be cool in ch-tools to know which shard the host belongs to. Do we have such an opportunity? Using this the code can be made much better

Shard name can be obtained from ClickHouse config, "shard" macro. Another option is to add a setting with shard name to ch-tools config.

Alex-Burmak (approved):

NikitaUnisikhin:
I'm inclined to leave the current implementation for flexibility in storing statistics in zk. 
@Alex-Burmak I don't have rights to merge. Merge please if there are no more comments on this PR.

PR #229, opened 2024-08-23
Introduce object storage cleaning on the shard scope (by default)

aalexfvk:
- Introduce object storage cleaning on the shard scope (by default) using `remote/remoteSecure` table functions.
- Add new parameter `--clean-scope` for `object-storage clean` command with possible values `host`, `shard`, `cluster`. `shard` is used by default.
- Deprecate `--on-cluster` parameter. Now it is not respected anymore. Add hint about this in the help message.

### DIFF
--- a/ch_tools/chadmin/cli/object_storage_group.py
+++ b/ch_tools/chadmin/cli/object_storage_group.py
@@ -78,12 +78,6 @@ def object_storage_group(ctx: Context, disk_name: str) -> None:
     help=("End of inspecting interval in human-friendly format."),
 )
 @option(
-    "--on-cluster/--no-on-cluster",
-    "_on_cluster",
-    is_flag=True,
-    help=("List objects on all hosts in a cluster. Deprecated, see --clean-scope"),
-)
-@option(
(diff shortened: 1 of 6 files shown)


ch_tools/chadmin/cli/object_storage_group.py
```diff
@@ -75,9 +79,16 @@ def object_storage_group(ctx: Context, disk_name: str) -> None:
 )
 @option(
     "--on-cluster/--no-on-cluster",
-    "on_cluster",
+    "_on_cluster",
     is_flag=True,
-    help=("List objects on all hosts in a cluster."),
+    help=("List objects on all hosts in a cluster. Deprecated, see --clean-scope"),
```
Alex-Burmak, line 84: I'd just drop it without deprecation :)

Alex-Burmak (approved):

ch_tools/chadmin/cli/object_storage_group.py
aalexfvk: I don't mind, done it

PR #230, opened 2024-09-05
Set receive_timeout for anti join query

aalexfvk:


ch_tools/common/commands/clean_object_storage.py
```diff
@@ -159,9 +159,11 @@ def _clean_object_storage(
     deleted = 0
     total_size = 0
     with TemporaryFile() as keys_file:
+        timeout = ctx.obj["config"]["object_storage"]["clean"]["antijoin_timeout"]
         with ch_client.query(
             antijoin_query,
-            timeout=ctx.obj["config"]["object_storage"]["clean"]["antijoin_timeout"],
+            timeout=timeout,
+            settings={"receive_timeout": timeout},
```
Alex-Burmak, line 166: To avoid similar issues with other queries, suggest to tune receive_timeout inside `query` based on `timeout` argument.

Alex-Burmak (approved):

PR #233, opened 2024-09-11
Update README and Makefile for macOS usage

NikitaUnisikhin:


Alex-Burmak (approved):

Makefile
```diff
@@ -78,10 +78,10 @@ export DEB_BUILD_DISTRIBUTION ?=
 
 
 define ensure_poetry
-	if [ ! -e $(POETRY) ]; then
-	  echo "Poetry could not be found. Please install it manually 'make install-poetry'";
-	  exit 1;
-	fi
+    if [ ! -e $(POETRY) ]; then \
```
aalexfvk, line 81: Pay attention on `.ONESHELL:` directive in the makefile. It enables to run multiline recepies in the same shell. It looks like this directive is not handled properly by your version of GNU Make. May be, it is old 3.81. (see [here](https://stackoverflow.com/questions/32153034/oneshell-not-working-properly-in-makefile) about it).
As an option, just updating of make utility would be solution for you. 
Alternatively, remove `.ONESHELL:` entirely and rewrite all recepies in the makefile for uniformity.

NikitaUnisikhin: Thanks, its work. Updated readme

PR #234, opened 2024-09-20
Kazoo auth for ch_keeper and keeper-monitoring

munakoiso:


ch_tools/common/clickhouse/config/clickhouse_keeper.py
```diff
@@ -38,6 +38,14 @@ def port_pair(self):
 
         return int(self._keeper_server.get("tcp_port", 0)), False
 
+    @property
+    def acl_username(self):
+        return self._keeper_server.get("acl_username")
```
Alex-Burmak, line 43: Does ClickHouse Keeper support this setting? Don't see it in https://clickhouse.com/docs/en/guides/sre/keeper/clickhouse-keeper#configuration

munakoiso: No
I thought there was only one configuration file for this tool, but there is another one.

ch_tools/monrun_checks_keeper/keeper_commands.py
```diff
@@ -26,15 +28,30 @@
 def alive_command(ctx):
     """Check (Zoo)Keeper service is alive"""
     try:
+        config = load_config()
```
Alex-Burmak, line 31: At this point the config should be already loaded and available in the context (`ctx.obj["config"]`).

Example of accessing config in another command:
https://github.com/yandex/ch-tools/blob/63be161b4a3fb3e45ef361100f8d25b0af92b5e2/ch_tools/chadmin/cli/object_storage_group.py#L45

Alex-Burmak (approved):

PR #247, opened 2024-10-31
Added error handle to orphaned objects monitoring

NikitaUnisikhin:


ch_tools/chadmin/cli/object_storage_group.py
```diff
@@ -195,3 +211,10 @@ def _table_formatter(stats):
     print_response(
         ctx, clean_stats, default_format="table", table_formatter=_table_formatter
     )
+
+
+def create_orphaned_objects_state(total_size: int, error_msg: str) -> dict:
```
aalexfvk, line 216: How about creating a dataclass for this purpose with methods `to_json()` and `from_json()` ?

NikitaUnisikhin: I agree, done

aalexfvk (approved):

PR #249, opened 2024-11-02
Not full work before reattach partition

k-morozov:


ch_tools/chadmin/cli/data_store_group.py
```diff
@@ -405,14 +379,42 @@ def find_paths_to_part_with_lost_keys(ctx: Context, root_path: str) -> List[str]
...
+    partition_list = get_partitions_by_path(ctx, parts_paths_with_lost_keys)
+    print_response(ctx, partition_list, default_format="table")
+
+    logging.debug(
+        "Found parts with missing s3 keys. Local paths of parts: {}",
+        parts_paths_with_lost_keys,
+    )
+
+
+def try_restore_data(ctx, partition_list, reattach, detach):
```
MikhailBurdukov, line 403: typing and doctstring

ch_tools/chadmin/cli/data_store_group.py
```diff
@@ -405,14 +379,42 @@ def find_paths_to_part_with_lost_keys(ctx: Context, root_path: str) -> List[str]
...
+    if not (reattach and detach):
```
MikhailBurdukov, line 404: Check covers only `reattach=true ; detach=true`. Lest remove it because we already have checked this as cli opt. Or make it `not (reattach xor detach)`.

ch_tools/chadmin/cli/data_store_group.py
```diff
@@ -405,14 +379,42 @@ def find_paths_to_part_with_lost_keys(ctx: Context, root_path: str) -> List[str]
                 s3_client, disk_conf.bucket_name, full_key
             ):
                 logging.debug("Not found key {}", full_key)
-                result.append(path)
+                parts_paths_with_lost_keys.append(path)
+
+                partition_list = get_partitions_by_path(ctx, [path])
+                try_restore_data(ctx, partition_list, reattach, detach)
```
MikhailBurdukov, line 385: check that we already handle the partition is needed. Because it doesn't make sense to run reattach/detach several time for the single partition.

ch_tools/chadmin/cli/data_store_group.py
```diff
@@ -350,33 +350,7 @@ def collect_orphaned_sql_objects_recursive(
...
-                    detach=True,
-                )
-
-
-def find_paths_to_part_with_lost_keys(ctx: Context, root_path: str) -> List[str]:
-    """
-    Find paths of parts with keys that doesn't have objects in s3.
-    """
-
-    result = []
```
MikhailBurdukov, line 379: Lets keep the main function small and move all logic to the another function.

ch_tools/chadmin/cli/data_store_group.py
k-morozov: Update

MikhailBurdukov (approved):

PR #252, opened 2024-11-05
Fix double detach

k-morozov:


ch_tools/chadmin/cli/data_store_group.py
```diff
@@ -402,7 +402,13 @@ def detect_broken_partitions(ctx, root_path, reattach, detach):
                     table_partition.table,
                     table_partition.partition,
                 )
-                try_repair_partition(ctx, table_partition, detach, reattach)
+                if detach:
+                    try_repair_partition(ctx, table_partition, False)
+                elif reattach:
+                    try_repair_partition(ctx, table_partition)
+                else:
+                    logging.debug("Skip repairing.")
```
MikhailBurdukov, line 410: This log msg are uninformative, lets remove it.

And log msg above:
```
Add part to check with path={}, table={}, partition={}
```
Lets make it:
```
Found the partition with missing blob in the object storage: path={} table={} partition={} 
```

k-morozov: ok

MikhailBurdukov (approved):

PR #287, opened 2025-02-06
Fix validate_version

Alex-Burmak:


Alex-Burmak:
The changes fix the issue from https://github.com/yandex/ch-tools/actions/runs/13172718693/job/36765818244

ch_tools/chadmin/internal/system.py
```diff
@@ -7,9 +7,9 @@
 
 
 def validate_version(version: str) -> None:
-    pattern = r"^\d{2}\.\d{2}\.\d+\.\d+$"
```
k-morozov, line 10: I see. The issue relates fort *.8.*.*
nit: we could just use `^\d{2}\.\d{1,2}\.\d+\.\d+$`

k-morozov (approved):

PR #289, opened 2025-02-13
Clean object storage without temporary file

kirillgarbar:


ch_tools/common/commands/clean_object_storage.py
```diff
@@ -163,33 +162,58 @@ def _clean_object_storage(
...
+        antijoin_query,
+        timeout=timeout,
+        settings={"receive_timeout": timeout, "max_execution_time": 0},
+        stream=True,
+        format_="TabSeparated",
+    ) as resp:
+        last_line = ""
+        # Setting chunk_size to 10MB, but usually incoming chunks are not larger than 1MB
+        for chunk in resp.iter_content(chunk_size=10 * 1024 * 1024):
+            keys, last_line = _split_response_to_keys(chunk, last_line)
```
aalexfvk, line 176: Do we need a custom function ? Have you considered [Response.iter_lines()](https://docs.python-requests.org/en/latest/api/#requests.Response.iter_lines) ?

kirillgarbar: I thought it will not work because the last line might break at the boundary of the chunk. I looked at the source code, it turns out it handles this case, so iter_lines can be used

PR #293, opened 2025-02-26
Add user and password for antijoin_query in clean object storage

NikitaUnisikhin:
Fix "Code: 516. DB::Exception: ... (AUTHENTICATION_FAILED)"

### DIFF
--- a/ch_tools/chadmin/cli/object_storage_group.py
+++ b/ch_tools/chadmin/cli/object_storage_group.py
@@ -27,6 +27,8 @@ STREAM_TIMEOUT = 10 * 60
 
 STATE_LOCAL_PATH = "/tmp/object_storage_cleanup_state.json"
 
+DEFAULT_CLUSTER_NAME = "{cluster}"
+
 
 @group("object-storage", cls=Chadmin)
 @option(
@@ -88,7 +90,7 @@ def object_storage_group(ctx: Context, disk_name: str) -> None:
 @option(
     "--cluster",
     "cluster_name",
-    default="{{cluster}}",
+    default=DEFAULT_CLUSTER_NAME,
     help=("Cluster to be cleaned. Default value is macro."),
 )
 @option(
@@ -143,6 +145,10 @@ def clean_command(
     deleted = 0
     total_size = 0
     error_msg = ""
+
+    # Need for correctly call format in Query
+    if cluster_name == DEFAULT_CLUSTER_NAME:
+        cluster_name = "{" + cluster_name + "}"
 
     try:
         deleted, total_size = clean(
--- a/ch_tools/chadmin/cli/object_storage_group.py
+++ b/ch_tools/chadmin/cli/object_storage_group.py
@@ -88,7 +88,7 @@ def object_storage_group(ctx: Context, disk_name: str) -> None:
 @option(
     "--cluster",
     "cluster_name",
-    default="{cluster}",
+    default="{{cluster}}",
     help=("Cluster to be cleaned. Default value is macro."),
 )
 @option(
--- a/ch_tools/common/clickhouse/client/query.py
+++ b/ch_tools/common/clickhouse/client/query.py
@@ -18,17 +18,19 @@ class Query:
     def _render(self, mask_sensitive: bool = True) -> str:
         if not self.sensitive_args:
             return self.value
-        if mask_sensitive:
-            sensitive_args = {key: self.mask for key in self.sensitive_args}
-        else:
-            sensitive_args = self.sensitive_args
+        sensitive_args = (
+            self._sensitive_args_mask() if mask_sensitive else self.sensitive_args
+        )
         return self.value.format(**sensitive_args)
+
+    def _sensitive_args_mask(self) -> Dict[str, str]:
+        return {key: self.mask for key in self.sensitive_args}
 
     def __str__(self) -> str:
         return self._render(True)
 
     def __repr__(self) -> str:
-        return f"{self.__class__.__name__}(value={self.value!r}, sensitive_args={self.sensitive_args!r})"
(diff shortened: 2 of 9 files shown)


tests/steps/common.py
```diff
@@ -134,3 +134,19 @@ def install_ch_tools_config_with_version(context, node):
...
+    password = ""
+
+    for node in nodes_list:
+        container = docker.get_container(context, node)
+        config_data = yaml.dump({"clickhouse": {"user": user, "password": password}})
+
+        container.exec_run(["bash", "-c", "mkdir /etc/clickhouse-tools"])
+        container.exec_run(
+            ["bash", "-c", f"echo '{config_data}' > /etc/clickhouse-tools/config.yaml"]
+        )
```
kirillgarbar, line 152: I think it is better to do some generic method that will set something in config. 
It can look like that
```
Scenario: Check clickhouse orphaned objects with state-zk-path option
    Given clickhouse-tools configuration
    """
        clickhouse:
            user: "_admin"
            password: ""
    """
    When we execute command on clickhouse01
    .....
```
`step_configuration(context):` is already working this way.

NikitaUnisikhin: done

NikitaUnisikhin:
Maybe we can wrap all queries in Query?

NikitaUnisikhin:
alternative: add an optional query_logging parameter to ch_client.query

ch_tools/common/clickhouse/client/query.py
```diff
@@ -0,0 +1,15 @@
...
+        self.value = value
+        self.sensitive_data = list(sensitive_data)
+
+    def for_execute(self) -> str:
+        return self.value
+
+    def for_logging(self) -> str:
+        result = self.value
+        for data in self.sensitive_data:
+            result = result.replace(data, self.mask)
```
aalexfvk, line 14: It looks a bit dirty, because can replace non-sensitive data in some circumstances.
How about passing value in the pre-formatted form of "... {username},  {password}... " and a dictionary of the sensitive data.
And use `value.format(**sensitive_dict)` instead of `str.replace()`

aalexfvk: And it makes sense to implemet `__str__` (instead of `for_logging()`), `__repr__` etc. 
You can see and example [here](https://github.com/pydantic/pydantic/blob/f6725ba21e4409cbd867427203adf85c39e42630/pydantic/types.py#L1541)

aalexfvk:
Please, add unit tests for `Query` class. I think it is good candidate for writing them.

ch_tools/common/clickhouse/client/query.py
NikitaUnisikhin: about the format, it will work incorrectly if such queries add values ​​that need to be replaced using format: https://github.com/yandex/ch-tools/blob/eb41df76e2b6a5ca1863bf59d2afd14969f5692d/tests/features/chadmin.feature#L14

NikitaUnisikhin: I agree with the other, I'll correct it.

NikitaUnisikhin: > about the format, it will work incorrectly if such queries add values ​​that need to be replaced using format:
> 
> https://github.com/yandex/ch-tools/blob/eb41df76e2b6a5ca1863bf59d2afd14969f5692d/tests/features/chadmin.feature#L14

`    # format function that does not throw an error if a substitution is not found
    def _format(self, args: dict) -> str:
        result = self.value
        for key, value in args.items():
            result = result.replace((key, value))
        return result`
I suggest adding a method where the key matches the '{*}' pattern (I can add validation to this in Query init)

NikitaUnisikhin:
For queries with sensitive data we use Query, for everything else - string, although Query is also possible. We rarely encounter queries with sensitive data, so I decided to leave 2 types of query representation.

NikitaUnisikhin:
https://github.com/yandex/ch-tools/blob/8d24b11ae6fd9bc144d948cfb2e837913b288051/ch_tools/common/clickhouse/client/clickhouse_client.py#L254
I decided not to move render_query from clickhouse_client to Query, since this requires knowing the version of CH, which is semantically not suitable for Query. It might be worth passing the CH version to render_query, but I think it looks better within the clickhouse_client framework.

ch_tools/common/clickhouse/client/query.py
```diff
@@ -0,0 +1,40 @@
...
+            sensitive_args = {key: self.mask for key in self.sensitive_args}
+        else:
+            sensitive_args = self.sensitive_args
+        return self.value.format(**sensitive_args)
+
+    def __str__(self) -> str:
+        return self._render(True)
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}(value={self.value!r}, sensitive_args={self.sensitive_args!r})"
```
aalexfvk, line 31: `__repr __()` also should return masked values. Only `for_execute()` returns original values

ch_tools/chadmin/cli/object_storage_group.py
```diff
@@ -88,7 +88,7 @@ def object_storage_group(ctx: Context, disk_name: str) -> None:
 @option(
     "--cluster",
     "cluster_name",
-    default="{cluster}",
+    default="{{cluster}}",
```
aalexfvk, line 91: It's not very pretty from a user perspective. Maybe double them in the code.

NikitaUnisikhin:
failed tests are not related to changes.

ch_tools/chadmin/cli/object_storage_group.py
NikitaUnisikhin: done

ch_tools/common/clickhouse/client/query.py
NikitaUnisikhin: done

aalexfvk (approved):

PR #301, opened 2025-03-14
Idempotent restore replica

kirillgarbar:


ch_tools/chadmin/internal/table_replica.py
```diff
@@ -168,3 +176,106 @@ def system_table_drop_replica(ctx, replica, database, table, dry_run=False):
...
+        restore_table_replica(
+            ctx,
+            database,
+            table,
+            cluster=cluster,
+            dry_run=dry_run,
+        )
+    except ClickhouseError as e:
+        msg = e.response.text.strip()
+        if "BAD_ARGUMENTS" in msg:
```
aalexfvk, line 227: Please write an example of the case when we get `BAD_ARGUMENTS` for history

ch_tools/chadmin/internal/table_replica.py
```diff
@@ -168,3 +176,106 @@ def system_table_drop_replica(ctx, replica, database, table, dry_run=False):
...
+    return not bool(check_zk_node(ctx, replicas_path))
+
+
+def table_is_readonly(ctx, database_name, table_name):
+    return bool(get_table_replica(ctx, database_name, table_name)["is_readonly"])
+
+
+def restore_replica(ctx, database, table, cluster, dry_run):
+    # TODO: remove attach when restore is fixed in ClickHouse
+    is_first_replica = no_replicas_in_zookeeper(ctx, database, table)
```
aalexfvk, line 213: Actually we can get a race a condition here between `no_replicas_in_zookeeper` on different replicas.

But we can check that this replica is first by comparing `czxid` of `../replicas` node and `../replicas/current_replica_hostname`. If they are equal we are the first and should attach our parts.

aalexfvk:
And rebase pls to make tests green

aalexfvk (approved):

PR #306, opened 2025-03-28
zero-copy migration script

MikhailBurdukov:
Im pretty sure that the logic will be never merged but it looks like useful script to migrate cluster from one zk path to another one.

another great feature that it is resumable, so if smth goes wrong you can resume the execution.

## Summary by Sourcery

Add a zero-copy migration script to the chadmin CLI that supports detaching tables, cleaning ZooKeeper nodes, restarting ClickHouse, and restoring replicas in parallel, with progress persisted to a status file for resumability.

New Features:
- Add a zero-copy migration CLI group with a resumable “migrate” command for migrating replicated tables between ZooKeeper paths
- Persist migration progress in a YAML status file to allow interruption and resumption of the migration workflow

Enhancements:
- Extend execute_tasks_in_parallel to accept an optional callback for tracking task completion and updating migration status


Alex-Burmak:
@sourcery-ai review

PR #310, opened 2025-04-13
Suppresse orphaned-objects monrun check if clickhouse uptime is low

NikitaUnisikhin:


ch_tools/monrun_checks/ch_orphaned_objects.py
```diff
@@ -48,6 +51,9 @@ def orphaned_objects_command(
     if not ClickhouseConfig.load().storage_configuration.has_disk("object_storage"):
         return Result(OK, "Disabled")
 
+    if get_uptime(ctx) == timedelta():
```
Alex-Burmak, line 54: Suggest to have some threshold on min uptime as in the backup check. It should be big enough (something like 1 day) because `ch-monitoring orphaned-objects` just report status from the state file and actual search of orphan objects is performed outside of this command.

NikitaUnisikhin: I agree, thank you, I corrected it

PR #314, opened 2025-04-28
check table schema

k-morozov:


ch_tools/chadmin/internal/migration.py
```diff
@@ -78,6 +82,28 @@ def migrate_as_non_first_replica(ctx, database_name, temp_db):
     original_engine = metadata_non_repl_db.database_engine
     tables_info = _get_tables_info_and_detach(ctx, database_name)
 
+    # Unfortunately, it is not atomic.
+    # The schema could be changed before detach databases.
+    # However, it is sufficient for now, and we will stabilize
+    # this logic after we stop creating temporary tables.
+    for row in tables_info:
```
k-morozov, line 89: We could not check the schema after detaching the database, because if the schema is different, we would not be able to reattach the temporary database (for example, on the second replica).

ch_tools/chadmin/internal/migration.py
```diff
@@ -285,6 +311,35 @@ def _get_tables_info_and_detach(ctx: Context, database_name: str) -> dict:
...
+    zk_metadata_path = f"/clickhouse/{database_name}/metadata/{table_name}"
+    zk_table_metadata = get_zk_node(ctx, zk_metadata_path)
+
+    local_table_metadata = read_local_table_metadata(ctx, table_local_metadata_path)
+
+    zk_table_metadata = zk_table_metadata.rstrip()
+    local_table_metadata = local_table_metadata.rstrip()
+
+    # Do not need use regex because we have stable table metadata format
+    start_pos_uuid = 21
```
aalexfvk, line 326: Maybe use string constants ?

ch_tools/chadmin/internal/migration.py
```diff
@@ -78,6 +82,28 @@ def migrate_as_non_first_replica(ctx, database_name, temp_db):
     original_engine = metadata_non_repl_db.database_engine
     tables_info = _get_tables_info_and_detach(ctx, database_name)
 
+    # Unfortunately, it is not atomic.
+    # The schema could be changed before detach databases.
+    # However, it is sufficient for now, and we will stabilize
+    # this logic after we stop creating temporary tables.
+    for row in tables_info:
+        if "Replicated" in row["engine"]:
+            logging.info("Table engine {} can have different schema", row["engine"])
```
aalexfvk, line 91: It makes sense to check all engines. Maybe show warning and continue process for replicated engines

k-morozov: Warning was added.

Alex-Burmak:
@sourcery-ai review

PR #315, opened 2025-04-28
add disallow_untyped_defs to mypy 

k-morozov:


ch_tools/common/clickhouse/client/utils.py
```diff
@@ -1,4 +1,4 @@
-def _format_str_match(value):
+def _format_str_match(value: str) -> str:
```
Alex-Burmak, line 1: Shouldn't it be 

```
def _format_str_match(value: Optional[str]) -> Optional[str]:
```

?

Alex-Burmak:
@sourcery-ai review

ch_tools/common/result.py
```diff
@@ -15,10 +18,10 @@ def __init__(self, code=OK, message="OK", verbose=""):
 class Status:
     """Class for holding Juggler status."""
 
-    def __init__(self):
+    def __init__(self) -> None:
         self.code = 0
-        self.text = []
-        self.verbose = []
+        self.text: List[str] = []
```
Alex-Burmak, line 23: `List[str]` is deprecated in favor of `list[str]`

https://docs.python.org/3/library/typing.html#typing.List

ch_tools/common/clickhouse/client/utils.py
k-morozov: Updated

ch_tools/common/result.py
k-morozov: Fixed

Alex-Burmak (approved):

PR #317, opened 2025-05-03
Integration with project and package management tool "uv"

Alex-Burmak:
https://docs.astral.sh/uv/

Peer PR in ch-backup repo: https://github.com/yandex/ch-backup/pull/227

PR #319, opened 2025-05-07
Fix 3.12 CI and sync git_igonore with make variable

k-morozov:
Make file contains `VENV_DIR := .venv`. However gitignore contains env. I assume it is confusing and we should use one value.

However, The CI with python3.12 was broken. 
Links:
- https://github.com/pypa/virtualenv/issues/2487
- https://github.com/python-poetry/poetry/issues/10382
- https://github.com/python-poetry/poetry/issues/10378

So I also have prepared the fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,7 @@ install-poetry:
 	# while 'poetry publish' for version poetry 1.1.15 due to incompatibility with urllib3 >= 2.0.0
 	# https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html#importerror-cannot-import-name-gaecontrib-from-requests-toolbelt-compat
 	$(POETRY_HOME)/venv/bin/python -m pip install "urllib3<2.0.0"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -136,6 +136,7 @@ install-poetry:
 	# while 'poetry publish' for version poetry 1.1.15 due to incompatibility with urllib3 >= 2.0.0
 	# https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html#importerror-cannot-import-name-gaecontrib-from-requests-toolbelt-compat
 	$(POETRY_HOME)/venv/bin/python -m pip install "urllib3<2.0.0"
+	$(POETRY_HOME)/venv/bin/python -m pip install "virtualenv>=20.30.0,<20.31.0"
```
Alex-Burmak, line 139: BTW in https://github.com/yandex/ch-tools/pull/317 support of Python 3.12 doesn't require special adjustment of venv. But it completely replaces poetry with another tool.

k-morozov: Seems that we just could wait #317 .

aalexfvk (approved):

PR #344, opened 2025-05-26
Set port to replicas from config

k-morozov:


k-morozov:
@sourcery-ai review

ch_tools/chadmin/internal/migration.py
```diff
@@ -196,7 +196,11 @@ def _get_host_id(ctx: Context, migrating_database: str, replica: str) -> str:
     rows = execute_query(ctx, query, echo=True, format_=OutputFormat.JSON)
     database_uuid = rows["data"][0]["uuid"]
 
-    result = f"{escape_for_zookeeper(host_name)}:9000:{database_uuid}"
+    query = "SELECT tcpPort() AS tcp_port"
```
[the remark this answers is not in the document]
k-morozov: Ok, will check later.

aalexfvk, line 199: Can we extract it from there https://github.com/yandex/ch-tools/blob/21259169a324eb05b8b9f67060677fdf041cf088/ch_tools/common/clickhouse/client/clickhouse_client.py#L289 ?

k-morozov: Yea, we could use port from config instead query. Updated

PR #354, opened 2025-05-29
Sync uuid for table_shared_id

k-morozov:


ch_tools/chadmin/cli/table_group.py
```diff
@@ -922,9 +922,8 @@ def change_uuid_command(
 
                 metadata = parse_table_metadata(table_local_metadata_path)
                 if not metadata.table_engine.is_table_engine_replicated():
-                    raise RuntimeError(
-                        f"Table {table_name} is not replicated. Failed get uuid from table_shared_id node."
-                    )
+                    logging.info(f"Table {table_name} is not replicated.")
```
aalexfvk, line 925:
```suggestion
                    logging.info(f"Table {table_name} is not replicated. Skip it.")
```

ch_tools/chadmin/cli/zookeeper_group.py
```diff
@@ -365,3 +368,69 @@ def remove_hosts_from_table(
...
+)
+@option(
+    "-dst",
+    "--dst",
+    type=ListParamType(),
+    required=True,
+    help="Path to the shard or shards in which the table nodes must be updated according to the source table_shared_id node",
+)
+@pass_context
+def sync_tables_nodes(ctx: Context, src: str, dst: list) -> None:
```
aalexfvk, line 401:
```suggestion
def sync_tables_nodes(ctx: Context, src: str, dst: list[str]) -> None:
```

ch_tools/chadmin/cli/zookeeper_group.py
```diff
@@ -365,3 +368,69 @@ def remove_hosts_from_table(
...
+                )
+
+                for zk_shard_path in dst:
+                    dst_shared_table_id_path = (
+                        f"{zk_shard_path}/{src_table_name}/table_shared_id"
+                    )
+                    txn.set_data(
+                        path=dst_shared_table_id_path, value=table_shared_id.encode()
+                    )
+                    txn.check(path=src_shared_table_id_path, version=stat.version)
```
aalexfvk, line 427: Is it added multiple times ?

ch_tools/chadmin/internal/zookeeper.py
```diff
@@ -19,6 +19,13 @@ def get_zk_node(ctx, path, binary=False):
         return value if binary else value.decode().strip()
 
 
+def get_zk_node_with_version(ctx, path, binary=False):
```
aalexfvk, line 22: Let's use type hints for new code

aalexfvk: How about passing client here ?

tests/features/chadmin_zookeeper.feature
```diff
@@ -198,3 +198,77 @@ Feature: chadmin zookeeper commands.
...
+    """
+    Then it completes successfully
+    When we execute command on clickhouse01
+    """
+    chadmin zookeeper update '/clickhouse/foo/shard1/test_table_2/table_shared_id' 'a7434a27-e2e5-41d1-be33-6a1e19b33ab3'
+    """
+    Then it completes successfully
+    When we execute command on clickhouse01
+    """
+    chadmin zookeeper sync-tables-nodes --src '/clickhouse/foo/shard1' --dst '/clickhouse/foo/shard2'
```
aalexfvk, line 256: Let's check list version also.
BTW, we can create different table replicas on the one host

tests/features/chadmin_zookeeper.feature
```diff
@@ -198,3 +198,77 @@ Feature: chadmin zookeeper commands.
...
+    a7434a27-e2e5-41d1-be33-6a1e19b33ab4
+    """
+    When we execute command on clickhouse02
+    """
+    chadmin zookeeper get '/clickhouse/foo/shard1/test_table_2/table_shared_id'
+    """
+    Then we get response
+    """
+    a7434a27-e2e5-41d1-be33-6a1e19b33ab3
+    """
```
aalexfvk, line 274: New line

aalexfvk (commented):
Should we do restart replica after this manipulations (maybe under option) ?

k-morozov:
> Should we do restart replica after this manipulations (maybe under option) ?

No.

tests/features/chadmin_zookeeper.feature
aalexfvk: I mean that `dst` can accept list of shards

k-morozov:
> I came across serious blocker, table_shared_id is used for composing zero-copy locks paths. And if we change shared_id for the table it will not find previously created locks, since they will be on old path. This will lead to data loss.

I saw the same thing in the logs after manually changing `table_shared_id` node as well. I'll add it later.

PR #357, opened 2025-06-06
database restore replica

k-morozov:


ch_tools/chadmin/cli/database_group.py
```diff
@@ -215,3 +217,47 @@ def migrate_engine_command(ctx: Context, database: str) -> None:
...
+
+
+@database_group.command("restore-replica")
+@option("-d", "--database", required=True)
+@pass_context
+def restore_replica_command(ctx: Context, database: str) -> None:
+    try:
+        if not is_database_exists(ctx, database):
+            logging.error("Database {} does not exists, skip restore", database)
+            sys.exit(1)
```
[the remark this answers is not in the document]
k-morozov: If we throw an exception chadmin  would be completed successfully.

For example, we need it for:
```
Then it fails with response contains
```

ch_tools/chadmin/cli/database_group.py
```diff
@@ -215,3 +217,47 @@ def migrate_engine_command(ctx: Context, database: str) -> None:
...
+
+        first_replica = True
+        try:
+            create_database_nodes(
+                ctx,
+                database_name=database,
+                db_replica_path=db_metadata.replica_path,
+            )
+        except NodeExistsError as ex:
```
MikhailBurdukov, line 244: IMO better to move it inside `restore_replica`. WDYT?

k-morozov: We use `restore_replica` in `migrate_as_first_replica` as well and we use the result (exception) of `create_database_nodes`  to specify the kind of migration and restore (it is first replica or no). I suggest separating logic for restore replica with checking is it first replica (`create_database_nodes`).

PR #373, opened 2025-07-02
Drop orphaned ZK paths for listing_objects_ table

MedvedewEM:
The problem appears if we move ('resetup") CH host without tables `listing_objects_...` or `orphaned_objects_`.
Then next time while `chadmin object-storage clean` command we have error: `REPLICA_ALREADY_EXISTS`, because it has ZK path, but table does not exists on host locally.

## Summary by Sourcery

Prevent `REPLICA_ALREADY_EXISTS` errors by dropping orphaned ZooKeeper replica paths for listing_objects_ and orphaned_objects_ tables before table operations.

Bug Fixes:
- Remove stale replica ZK nodes when the corresponding table is absent on a host to avoid replica-exists errors.

Enhancements:
- Introduce `try_query_json_data_first_row` helper to safely fetch an optional first row.
- Refactor ClickHouse client creation into `_clickhouse_client` and add `custom_clickhouse_client` to override host.
- Implement `_clean_orphaned_zk_paths` and invoke it during object listing table setup.

Tests:
- Add a BDD scenario to verify that orphaned ZK paths are cleaned in dry-run mode.

### DIFF
--- a/ch_tools/common/commands/clean_object_storage.py
+++ b/ch_tools/common/commands/clean_object_storage.py
@@ -7,6 +7,7 @@ from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple
 import click
 from click import Context
 from humanfriendly import format_size
+from kazoo.exceptions import NoNodeError
 
 from ch_tools.chadmin.internal.object_storage import cleanup_s3_object_storage
 from ch_tools.chadmin.internal.object_storage.obj_list_item import ObjListItem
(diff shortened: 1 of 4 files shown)


ch_tools/common/commands/clean_object_storage.py
```diff
@@ -51,13 +57,49 @@ class CleanScope(str, Enum):
...
+
+        ch_client = custom_clickhouse_client(ctx, host=replica)
+        res = ch_client.try_query_json_data_first_row(
+            query=f"SELECT name FROM system.tables WHERE database = '{database}' AND name = '{table}'",
+            compact=True,
+        )
+
+        if not res:
+            zk_replicas_path, _ = os.path.split(replicas_path)
+            system_table_drop_replica_by_zk_path(ctx, replica, zk_replicas_path)
```
MedvedewEM, line 91: There is possible data race if you run `chadmin object-storage clean` command on multiple hosts of the same shard.
But usually you should not run multiple `clean` commands on different replicas of one shard (or you should use lock mechanism on your own).

aalexfvk:
Check please why `cleanup-removed-hosts-metadata` does not work for the root cause. Should it do ?

MedvedewEM:
Dedided to clean such ZK paths outside of clickhouse-tools. Discarded

PR #390, opened 2025-09-10
Skip sanity-check for orphaned objects' cleaning in dry-run mode

MedvedewEM:
## Summary by Sourcery

Allow dry-run mode to bypass fatal sanity-check failures by logging warnings instead of raising errors and add tests for dry-run behavior.

Enhancements:
- Propagate the dry_run flag into the sanity-check function and convert errors into warnings when running in dry-run mode
- Skip the size threshold check altogether when the object bucket listing is empty

Tests:
- Update BDD tests to validate dry-run responses include WouldDelete and TotalSize and adjust command invocations and formatting
- Add a new scenario to confirm dry-run returns zero deletions and size when no objects are present

### DIFF
--- a/ch_tools/common/commands/clean_object_storage.py
+++ b/ch_tools/common/commands/clean_object_storage.py
@@ -238,13 +238,14 @@ def _sanity_check_before_cleanup(
                 )
 
     def perform_check_size() -> None:
+
+        if listing_size_in_bucket == 0:
+            return
+
         ### Total size of objects after cleanup must be very close to sum(bytes) FROM system.remote_data_paths
         size_to_delete = 0
         for orphaned_object in orphaned_objects_iterator():
             size_to_delete += orphaned_object.size
-
-        if listing_size_in_bucket == 0:
-            return
 
         if (
             listing_size_in_bucket * size_error_rate_threshold_fraction


MedvedewEM:
> Why we should? As for me it is good to check before the real run and also we can disable checks in the config if needed.

FYI I added new changes. Actually I am solving two problems now:

1) When I use `dry-run` option I would like to see the result (size of orphaned objects) in normal format, but not exception about dangerous operation (with exit-code == 1).
It is optional, I can get rid of these changes - solution with disable sanity-check with overriding setting - good for me.
2) If `listing_size_in_bucket == 0` (including newly created cluster), result should be successed anyway, but not like that:
```
(newly created cluster with s3)
# chadmin object-storage clean --dry-run
...........
    raise RuntimeError(
RuntimeError: Potentially dangerous operation: Going to remove more than 90% of bucket content. To remove 0 bytes; listing size 0 bytes
#
#
# echo $?
1
```

MikhailBurdukov:
>When I use dry-run option I would like to see the result (size of orphaned objects) in normal format, but not exception about dangerous operation (with exit-code == 1).
It is optional, I can get rid of these changes - solution with disable sanity-check with overriding setting when I use dry-run - good for me.

Hm. Just thoughts: or we can write warn msg into the log instead of throwing the exception; 

> If listing_size_in_bucket == 0 (including newly created cluster), result should be successed anyway, but not like that:

:+1:

PR #394, opened 2025-09-17
Analyze s3 space usage and collect orphaned objects in the same command

kirillgarbar:


ch_tools/common/commands/object_storage.py
```diff
@@ -102,43 +217,149 @@ def clean(
...
+    timeout = ctx.obj["config"]["object_storage"]["clean"]["antijoin_timeout"]
+    query_settings = {"receive_timeout": timeout, "max_execution_time": 0}
+    execute_query(
+        ctx,
+        space_usage_query,
+        timeout=timeout,
+        settings=query_settings,
+    )
+
+    execute_query(ctx, f"TRUNCATE TABLE {space_usage_table} SYNC")
```
aalexfvk, line 358: Why not `EXCHANGE TABLES` ?

kirillgarbar: After exchange their ZK paths will not be swapped, and `space_usage_table` will have `space_usage_table_new` path in ZK. It makes it impossible to recreate `space_usage_table_new` with the same path again, so we will have to keep it existing. Also, exchange on shard will not be atomic anyway, and delete + insert of one row will be pretty fast.

PR #396, opened 2025-09-17
Retries for system sync replica queries

MikhailBurdukov:
## Summary by Sourcery

Provide a retry mechanism and deadline handling for SYSTEM SYNC REPLICA queries in the chadmin CLI

New Features:
- Introduce execute_query_with_retries utility to wrap execute_query with retry logic and optional deadline
- Add sync-query-max-retries command-line option to configure retry attempts for replication sync

Enhancements:
- Update wait_replication_sync_command to use execute_query_with_retries and apply custom retriable error filtering
- Define CLICKHOUSE_TIMEOUT_EXCEEDED_MSG constant and streamline exception handling for sync queries

### DIFF
--- a/ch_tools/chadmin/cli/wait_group.py
+++ b/ch_tools/chadmin/cli/wait_group.py
@@ -4,6 +4,13 @@ from typing import Any, Optional
 
 import requests
 from click import Context, FloatRange, group, option, pass_context
+from tenacity import (
+    retry,
+    retry_if_exception_type,
+    retry_if_not_exception_message,
+    stop_after_attempt,
+    stop_after_delay,
+)
 
 from ch_tools.chadmin.cli.chadmin_group import Chadmin
 from ch_tools.chadmin.internal.clickhouse_disks import S3_METADATA_STORE_PATH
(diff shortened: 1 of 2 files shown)


aalexfvk:
It looks too complex. How about something like this:
```python
# utils.py
def execute_query(
    ctx: Context,
    ...
    settings: Optional[Any] = None,
    deadline: float = 0,
    **kwargs: Any,
) -> Any:
    """
    Execute ClickHouse query.
    """
    ...    
    if deadline:
        ts = time.time()
        if ts >= deadline:
            raise RuntimeError("Deadline reached")        
        settings = settings or {}
        settings["receive_timeout"] = deadline - ts

    return ch_client.query(
    ...
    )
```

```python
# wait_group.py
retry_decorator = tenacity.retry(
    retry=tenacity.retry_if_exception_type(ClickhouseError)
    & tenacity.retry_if_not_exception_message(match=rf".*TIMEOUT_EXCEEDED.*"),
    stop=tenacity.stop_after_attempt(max_attempts)
    reraise=True,
)
retry_decorator(execute_query)(
    ctx,
    query,
    format_=None,
    timeout=replica_timeout.total_seconds(),
    deadline=deadline
)
```

MikhailBurdukov:
To be honest I would prefer to isolate retry logic to separate function as much as possible; And actually for me `deadline ` looks too specific for such general function as `execute_query`, because we already have `timeout`. 
But of course it makes sense due to minimal changes and simplicity. Will do as you suggested, you are the gatekeeper   :smile:

PR #461, opened 2026-03-31
Attach partitions only if there are parts without prefixes

kirillgarbar:
## Summary by Sourcery

Add support for filtering and attaching partitions based on a textual reason associated with detached partitions.

New Features:
- Allow partition queries to be filtered by an optional reason field when retrieving partitions.

Enhancements:
- Propagate an optional reason parameter through partition retrieval and attach commands to support more precise partition management.

### DIFF
--- a/ch_tools/chadmin/cli/partition_group.py
+++ b/ch_tools/chadmin/cli/partition_group.py
@@ -214,6 +214,7 @@ def attach_partitions_command(
         table,
         partition_id=partition_id,
         detached=True,
+        reason="",
         format_="JSON",
         **kwargs,
     )["data"]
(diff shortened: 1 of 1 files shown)


ch_tools/chadmin/cli/partition_group.py
```diff
@@ -1048,6 +1050,9 @@ def get_partitions(
             {% if table -%}
               AND table {{ format_str_match(table) }}
             {% endif -%}
+            {% if reason is not none -%}
+              AND reason {{ format_str_match(reason) }}
+            {% endif -%}
```
kirillgarbar, line 1055: This behaviour is already present for `chadmin part attach` and I also tested it locally so I think it doesn't need an integration test

aalexfvk (approved):

PR #468, opened 2026-05-05
Add tcp mem check

ruMoonGuard:
## Summary by Sourcery

Add a monrun check for TCP memory usage based on /proc/net/sockstat and register it in the ch-monitoring CLI.

New Features:
- Introduce a tcp-mem monrun check command to report TCP memory usage and evaluate it against configurable warning and critical thresholds.

Tests:
- Add feature tests covering the tcp-mem check for default thresholds and critical threshold behavior.

### DIFF
--- a/tests/features/monrun.feature
+++ b/tests/features/monrun.feature
@@ -583,3 +583,23 @@ Feature: ch-monitoring tool
     """
     2;Metric not available
     """
+
+  Scenario: Check TCP memory sockstat with default thresholds
+    When we execute command on clickhouse01
+    """
+    ch-monitoring tcp-mem
+    """
+    Then we get response
+    """
+    0;OK
+    """
+
+  Scenario: Check TCP memory sockstat with critical threshold
+    When we execute command on clickhouse01
+    """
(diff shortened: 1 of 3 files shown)


tests/features/monrun.feature
```diff
@@ -583,3 +583,23 @@ Feature: ch-monitoring tool
...
+    """
+
+  Scenario: Check TCP memory sockstat with critical threshold
+    When we execute command on clickhouse01
+    """
+    ch-monitoring tcp-mem -w 0 -c -1
+    """
+    Then we get response contains
+    """
+    2;tcp_mem_bytes, crit = -1, count =
```
MedvedewEM, line 604: Why it is `count =`, but not `count = 0`?

ruMoonGuard: The exact `count` value is intentionally not asserted here. This test is focused on the threshold/check result, not on the specific count value. The count can vary between environments, so checking only `count =` makes the test more stable.

MedvedewEM (approved):

PR #490, opened 2026-09-07
MDB-43421: Fix silently dropped filters from partition queries

roaldm153:
**Problem:** 
+ The disk filter was ignored by `chadmin partition attach`, as it was never rendered in the query. 
+ Also, two more filters, `--min/max-part` and `--min/max size`, were dropped in the same way.

**My solution:** 
+ Add required filters to the having section. 
+ Use `{{ min_size }}` instead of `'{{ min_size }}'` in the query text (same for `max_size`) - these fields should be compared as numbers, not strings. 
+ Add unit tests to cover these scenarios.
+ Add description for `min/max_partition` options.

## Summary by Sourcery

Preserve all requested filters when selecting partitions for attach operations.

Bug Fixes:
- Ensure partition attach filters are applied for disk, partition ID ranges, size, and part-count constraints.
- Correct size filtering to use numeric comparisons and avoid unsupported detached-part columns on older ClickHouse versions.

Enhancements:
- Add descriptions for minimum and maximum partition CLI options.

Tests:
- Add unit and feature coverage for disk, size, and part-count partition filtering, including ClickHouse version compatibility.


aalexfvk:
Failed tests are fixed here https://github.com/yandex/ch-tools/pull/491

otselnik:
@roaldm153 Что думаешь на счет добавления интеграционного теста?

aalexfvk:
Yes, let's add integration test to `tests/features/chadmin_partitions.feature`

aalexfvk (approved):

